PR #17, opened 2017-09-04
Add ActionController::Metal as possible base class of ApplicationController

aCandidMind:
This is making `rails g jsonapi_suite:install` more flexible with regard to the base class of the app's `ApplicationController`.

This is common when performance is deeply cared for. One use-case is
`ActionController::API.without_modules(blacklist)` to blacklist what's not needed.

The diff appears bigger than it is, I've just wrapped the old code in a loop.


richmolj:
I like it - thanks!

aCandidMind:
Thank you for open sourcing your gems, the documentation and all that! And for the fast merge.

PR #21, opened 2017-11-21
Fix typo in allow_sideload example

Subtletree:


richmolj:
@Subtletree whoops, thanks for the edit! I think you want to edit `concepts.md` and then run `bundle exec jekyll build` or `bundle exec jekyll serve` to build the compiled files though.

Subtletree:
struggling 😅

richmolj:
Somehow I don't see a change to `concepts.md` in the diff?

Subtletree:
Think `concepts.md` just iterates over each concept. So the change is in `_includes/docs/sideloading/example.html`.

richmolj:
Ah, of course. Thanks again!

PR #26, opened 2018-02-27
Fix broken link to side-effects

stevenharman:


richmolj:
Much thanks! I think you’ll have to “jekyll build” to actually fix the site though.

stevenharman:
Doh! My mistake. I was assuming this was a GitHub Pages build. Will do.

richmolj:
Looks like the link is now working!

(I actually need to update these to point to the [new docs](https://jsonapi-suite.github.io/jsonapi_suite/ruby/) anyway, so thanks for the reminder :)

PR #29, opened 2018-03-02
Show jsonapi-compliant errors on invalid param

wadetandy:
Previously, when requests were made that violated the formats being
checked by strong_resources, there was an error thrown which resulted
in a generic 500 error being sent back to the user, which wasn't
particularly helpful.  Those errors are now caught, parsed, and returned
as a jsonapi-compliant error object with a 400 (bad request) status.

### DIFF
--- a/lib/jsonapi_suite.rb
+++ b/lib/jsonapi_suite.rb
@@ -5,6 +5,8 @@ require 'jsonapi_compliable'
 require 'jsonapi_errorable'
 
 require "jsonapi_suite/version"
+require "jsonapi_suite/strong_resources/exception_handler"
+require "jsonapi_suite/strong_resources/controller_mixin"
 require "jsonapi_suite/controller_mixin"
 
 module JsonapiSuite
--- a/lib/jsonapi_suite/strong_resources/exception_handler.rb
+++ b/lib/jsonapi_suite/strong_resources/exception_handler.rb
@@ -1,0 +1,37 @@
+module JsonapiSuite
+  module StrongResources
+    class ExceptionHandler < JsonapiErrorable::ExceptionHandler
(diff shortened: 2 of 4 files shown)


lib/jsonapi_suite/strong_resources/exception_handler.rb
```diff
@@ -0,0 +1,37 @@
+module JsonapiSuite
+  module StrongResources
+    class ExceptionHandler < JsonapiErrorable::ExceptionHandler
```
richmolj: Can this class live within SR itself? I am thinking it might be more testable/maintainable long-term.

lib/jsonapi_suite/strong_resources/exception_handler.rb
```diff
@@ -0,0 +1,37 @@
...
+        {
+          "errors" => [error]
+        }
+      end
+
+      def status_code(_)
+        400
+      end
+
+      def log?
```
richmolj: I think we might want `true` here

lib/jsonapi_suite.rb
```diff
@@ -5,6 +5,8 @@
 require 'jsonapi_errorable'
 
 require "jsonapi_suite/version"
+require "jsonapi_suite/strong_resources/exception_handler"
+require "jsonapi_suite/strong_resources/controller_mixin"
```
richmolj: My main comment is I found the naming here really confusing until I dug into it. I'm thinking exception_handler can move to SR itself, and the mixin behavior could move into the original SR controller mixin? Any issues you see with that?

lib/jsonapi_suite/strong_resources/exception_handler.rb
wadetandy: I actually didn't know we were the owners of `SR`, which is why I did it here. That makes much more sense yes.

PR #31, opened 2018-04-09
BUGFIX - bad paths to QUICKSTART page

campbell:
Not sure why the `jsonapi_suite-0.7.0.gem` file was removed.


richmolj:
I'm hesitant to merge this code, for no other reason than I've tried it every which way and it often breaks in production. A lot of documentation and things that *seem* like they should work do not, which is really frustrating. I've been changing [this line](https://github.com/jsonapi-suite/jsonapi_suite/blob/gh-pages/_config.yml#L34) manually in development.

Will try merging later when I can be ready to revert immediately if something goes wrong. I **do** love the premise and would very much like to move past this hack, appreciate the help ❤️

PR #49, opened 2019-01-07
Add default_attributes class method to ApplicationResource

sideshowbandana:
If attributes aren't supplied for the graphiti:resource, it's possible
to derive the attributes from the existing schema for the model.

Resuming PR from: https://github.com/graphiti-api/graphiti_old_archived/pull/76


richmolj:
@sideshowbandana what were your thoughts on https://github.com/graphiti-api/graphiti_old_archived/pull/76#issuecomment-451664805?

wadetandy:
From the other thread:

> Another way to accomplish a similar purpose would be a generator that does the same, introspecting the DB and adding each attribute line. This way there's no additional DSL, and devs are a little more conscious about what they are opting into. What do you think about changing to this approach?

> Finally, this is ActiveRecord specific and we'll have to think about how to make that clear. Maybe another case for generators.

I generally agree with the above.  I think this is a really useful bootstrapping feature when adding a new resource for an existing model, but having non-explicit attributes on resources will become hard to manage and figure out what's going on.  I especially wouldn't want to inadvertently have users expose sensitive data because they didn't realize something was exposed by default.  Using this introspection to create attributes at generation time seems like a better compromise to me.

sideshowbandana:
I like the idea of using the generator to write the attributes into the resource. The tricky part is that the user will have to create the model and migrate the database before generating the resource in order for it to have the correct data. Also, if the user changes the migration or adds another migration, the generator will have to be run again (after the migrations are run) or the changes will have to be applied manually.

wadetandy:
> Also, if the user changes the migration or adds another migration, the generator will have to be run again (after the migrations are run) or the changes will have to be applied manually.

I think this is a feature, not a drawback.  Generators are there to get you up and running quickly, but at the end of the day the code that results is explicit and standard and up to the developer to evolve further.  If a resource is very simple, then it could be easier for the dev to regenerate it if the model changes, but otherwise they can explicitly add the additional fields they'd like to expose.

lib/generators/graphiti/templates/application_resource.rb.erb
```diff
@@ -12,4 +12,21 @@ class ApplicationResource < Graphiti::Resource
   self.adapter = Graphiti::Adapters::ActiveRecord
   self.base_url = Rails.application.routes.default_url_options[:host]
   self.endpoint_namespace = '<%= api_namespace %>'
+
+
+  def self.default_attributes(opts = { })
+    begin
+      unless opts[:from].kind_of?(Class) && opts[:from] <= ApplicationRecord
```
sideshowbandana: Should this be a string instead of a class? Similar to how ActiveRecord's `belongs_to` has a `class_name` option which has a string value.

richmolj:
@sideshowbandana I think I'm leaning toward the generator approach and agree with @wadetandy 's comment. What do you think of doing this with generators and omitting the class method?

sideshowbandana:
> @sideshowbandana I think I'm leaning toward the generator approach and agree with @wadetandy 's comment. What do you think of doing this with generators and omitting the class method?

Fair enough. See the recent patch set.

richmolj (commented):
Minor comments but let's get this merged soon!

lib/generators/graphiti/resource_generator.rb
```diff
@@ -21,9 +21,10 @@ class ResourceGenerator < ::Rails::Generators::NamedBase
       aliases: ['--actions', '-a'],
       desc: 'Array of controller actions to support, e.g. "index show destroy"'
 
-    class_option :'default-attributes-class',
+    class_option :'attributes-from',
       banner: 'Model',
       type: :string,
+      alias: ["--attributes-from", ""]
```
richmolj: How about `--model` and `-m`, to note we'll grab attributes by introspecting that model?

richmolj:
Quick note on the alias, otherwise LGTM 👍 !

richmolj:
@sideshowbandana does this work, currently? I see it referencing `attributes_class` and `default_attributes_class` and seems like the generator would blow up.

richmolj:
@sideshowbandana is this good to go now?

sideshowbandana:
I thought it was a while back.  Is there still an outstanding action item?

richmolj:
I think there were some comments in Slack and here that were addressed, just want to make sure it's done and tested. In particular I want to make sure we're running through the Quickstart and verifying everything still works.

sideshowbandana:
Ah right. Ok. I'll try to get to this during the week.

sideshowbandana:
Looks like the generator is working as expected. I followed the quick start and appended this to `all.rb`:
```
after_bundle do
  say(set_color("\nCreate Models
", :green, :bold))
  run("bundle exec rails generate model user name:string")
  run("bundle exec rails generate model device user:references name:string")
    say(set_color("\nMigrate
", :green, :bold))
  rails_command("db:migrate", env: 'development')
  rails_command("db:migrate", env: 'test')
  say(set_color("\nCreate Resources
", :green, :bold))
  run('bundle exec rails generate graphiti:resource device')
  run('bundle exec rails generate graphiti:resource device id:integer user_id:integer name:string created_at:datetime updated_at:datetime')
  run('bundle exec rails generate graphiti:resource device -m')
  run('bundle exec rails generate graphiti:resource device -m Device')
end
```

Here are the results:

```
Create Models

         run  bundle exec rails generate model user name:string from "."
Running via Spring preloader in process 30370
      invoke  active_record
      create    db/migrate/20190312073529_create_users.rb
      create    app/models/user.rb
      invoke    rspec
      create      spec/models/user_spec.rb
      invoke      factory_bot
      create        spec/factories/users.rb
         run  bundle exec rails generate model device user:references name:string from "."
Running via Spring preloader in process 30374
      invoke  active_record
      create    db/migrate/20190312073530_create_devices.rb
      create    app/models/device.rb
      invoke    rspec
      create      spec/models/device_spec.rb
      invoke      factory_bot
      create        spec/factories/devices.rb

Migrate

       rails  db:migrate
== 20190312073529 CreateUsers: migrating ======================================
-- create_table(:users)
   -> 0.0008s
== 20190312073529 CreateUsers: migrated (0.0009s) =============================

== 20190312073530 CreateDevices: migrating ====================================
-- create_table(:devices)
   -> 0.0015s
== 20190312073530 CreateDevices: migrated (0.0016s) ===========================

       rails  db:migrate
== 20190312073529 CreateUsers: migrating ======================================
-- create_table(:users)
   -> 0.0005s
== 20190312073529 CreateUsers: migrated (0.0005s) =============================

== 20190312073530 CreateDevices: migrating ====================================
-- create_table(:devices)
   -> 0.0009s
== 20190312073530 CreateDevices: migrated (0.0010s) ===========================


Create Resources

         run  bundle exec rails generate graphiti:resource device from "."
Running via Spring preloader in process 30393
      create  app/controllers/devices_controller.rb
      insert  config/routes.rb
      create  app/resources/device_resource.rb
      create  spec/resources/device/writes_spec.rb
      create  spec/resources/device/reads_spec.rb
      create  spec/api/v1/devices/index_spec.rb
      create  spec/api/v1/devices/show_spec.rb
      create  spec/api/v1/devices/create_spec.rb
      create  spec/api/v1/devices/update_spec.rb
      create  spec/api/v1/devices/destroy_spec.rb
         run  bundle exec rails generate graphiti:resource device id:integer user_id:integer name:string created_at:datetime updated_at:datetime from "."
Running via Spring preloader in process 30401
   identical  app/controllers/devices_controller.rb
      insert  config/routes.rb
    conflict  app/resources/device_resource.rb
Overwrite /Users/kyle/src/github.com/sideshowbandana/test_generators/blog/app/resources/device_resource.rb? (enter "h" for help) [Ynaqdhm] d
  class DeviceResource < ApplicationResource
+   attribute :id, :integer, writable: false
+   attribute :user_id, :integer
+   attribute :name, :string
+   attribute :created_at, :datetime, writable: false
+   attribute :updated_at, :datetime, writable: false
  end
Retrying...
Overwrite /Users/kyle/src/github.com/sideshowbandana/test_generators/blog/app/resources/device_resource.rb? (enter "h" for help) [Ynaqdhm] y
       force  app/resources/device_resource.rb
   identical  spec/resources/device/writes_spec.rb
   identical  spec/resources/device/reads_spec.rb
   identical  spec/api/v1/devices/index_spec.rb
   identical  spec/api/v1/devices/show_spec.rb
   identical  spec/api/v1/devices/create_spec.rb
   identical  spec/api/v1/devices/update_spec.rb
   identical  spec/api/v1/devices/destroy_spec.rb
         run  bundle exec rails generate graphiti:resource device -m from "."
Running via Spring preloader in process 30414
   identical  app/controllers/devices_controller.rb
      insert  config/routes.rb
   identical  app/resources/device_resource.rb
   identical  spec/resources/device/writes_spec.rb
   identical  spec/resources/device/reads_spec.rb
   identical  spec/api/v1/devices/index_spec.rb
   identical  spec/api/v1/devices/show_spec.rb
   identical  spec/api/v1/devices/create_spec.rb
   identical  spec/api/v1/devices/update_spec.rb
   identical  spec/api/v1/devices/destroy_spec.rb
         run  bundle exec rails generate graphiti:resource device -m Device from "."
Running via Spring preloader in process 30418
   identical  app/controllers/devices_controller.rb
      insert  config/routes.rb
   identical  app/resources/device_resource.rb
   identical  spec/resources/device/writes_spec.rb
   identical  spec/resources/device/reads_spec.rb
   identical  spec/api/v1/devices/index_spec.rb
   identical  spec/api/v1/devices/show_spec.rb
   identical  spec/api/v1/devices/create_spec.rb
   identical  spec/api/v1/devices/update_spec.rb
   identical  spec/api/v1/devices/destroy_spec.rb
```

richmolj:
Awesome! Looks great @sideshowbandana ! ❤️ 👏 Thanks for following up.

PR #51, opened 2019-01-07
Add pagination links

sideshowbandana:
Clients can specify if pagination links are desired. Developers can also enforce pagination links to be displayed by merging pagination_links: true into the params hash.

Resuming from PR: https://github.com/graphiti-api/graphiti_old_archived/pull/58


lib/graphiti/query.rb
```diff
@@ -30,6 +30,10 @@ def links?
       end
     end
 
+    def pagination_links?
+      return !!@params[:pagination_links]
+    end
```
richmolj, line 39: Could we add a global config for this as well? I am thinking we can do this like [Links](https://github.com/graphiti-api/graphiti/blob/1683c96774b6a9ab70e9e3c4b501efd3b3c2ac3d/lib/graphiti/query.rb#L24-L31). This way you can turn on pagination links by default, OR require the client to request them.

I think this is the **very last thing** for this PR and then let's merge.

richmolj (commented):
This is looking awesome ❤️ I have one last quick comment and then let's merge 🎉

richmolj:
Awesome! Thanks so much for your work on this and follow-through ❤️ 💯 🎉 !

richmolj:
@sideshowbandana one of my users had an issue with this right after release.

```
ERROR: NoMethodError: undefined method `/' for nil:NilClass
graphiti (1.0.rc.10) lib/graphiti/delegates/pagination.rb:40:in `last_page'
graphiti (1.0.rc.10) lib/graphiti/delegates/pagination.rb:15:in `block in links'
graphiti (1.0.rc.10) lib/graphiti/delegates/pagination.rb:13:in `tap'
graphiti (1.0.rc.10) lib/graphiti/delegates/pagination.rb:13:in `links'
graphiti (1.0.rc.10) lib/graphiti/renderer.rb:47:in `block in render'
graphiti (1.0.rc.10) lib/graphiti.rb:141:in `block in broadcast'
```

The larger issue is that we should have these links off by default. I did that [here](https://github.com/graphiti-api/graphiti/pull/81).

There also seemed to be some spec code that was problematic locally https://github.com/graphiti-api/graphiti/blob/47cea9c19731ae84ccf6112d66c54c69b5e67c41/spec/spec_helper.rb#L70-L73 - the merged code was `rspec` on line 70 and both seem to be in the wrong spot.

Could you look into these issues?

PR #52, opened 2019-01-07
Factory bot in generator

sideshowbandana:
Recreating PR from: https://github.com/graphiti-api/graphiti_old_archived/pull/74


richmolj:
@sideshowbandana I think the only thing to change is to use `attributes_for` instead of `build(...).attributes`?

sideshowbandana:
> @sideshowbandana I think the only thing to change is to use `attributes_for` instead of `build(...).attributes`?

`attributes_for` includes `id`, `updated_at`, and `created_at`. So, I'd have to generate code to strip the keys out.

sideshowbandana:
> > @sideshowbandana I think the only thing to change is to use `attributes_for` instead of `build(...).attributes`?
> 
> `attributes_for` includes `id`, `updated_at`, and `created_at`. So, I'd have to generate code to strip the keys out.

Ah, strike that. I forgot why I used `build` instead of `attributes_for`. Here's why:

Given a factory definition
```
FactoryBot.define do
  factory :my_model do
    other_model
    user { other_model.user }
    name { "MyString" }
  end
end
```
`attributes_for`, doesn't create the associations if necessary. In this case `other_model` Resulting in:
```
FactoryBot.attributes_for(:my_model)
NoMethodError: undefined method `user' for nil:NilClass
```
However, `build` will
```
FactoryBot.build(:my_model).attributes
=> {"id"=>nil, "user_id"=>1, "other_model_id"=>1, "name"=>"MyString", "created_at"=>nil, "updated_at"=>nil}
```

richmolj:
That seems like exactly the behavior we want, though? This test shouldn't include the associations in the `attributes` payload.

sideshowbandana:
How would we test the `./spec/resources/my_model/writes_spec.rb` in the case where the post needs to have the association ids due to foreign key constraints.

Is that on the user to generate the associated objects and manually add the ids into the payload?

Do you know of a better way to have a circular relationship (like this) in FactoryBot?
In other words, given:
```
FactoryBot.define do
  factory :my_model do
    other_model
    user { other_model.user }
    name { "MyString" }
  end
end
```
How can one achieve those relationships and call `attributes_for` without getting the exception:
```
FactoryBot.attributes_for(:my_model)
NoMethodError: undefined method `user' for nil:NilClass
```

sideshowbandana:
I've updated the generator to use attributes_for. I want to reiterate that there will be an issue with specs that infolve resources that have dependent associations.
Consider the following schema:
```
  create_table "users", force: :cascade do |t|
    t.string "email"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
  end
  create_table "posts", force: :cascade do |t|
    t.string "title"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.bigint "user_id"
    t.index ["user_id"], name: "index_posts_on_user_id"
  end
  add_foreign_key "posts", "users"
```
And the generated factories:
```
###
# bundle exec rails g model user email:string
# bundle exec rails g graphiti:resource user email:string
###
FactoryBot.define do
  factory :user do
    email { "MyString" }
  end
end
```
```
###
# rails g model post user:references
# rails g graphiti:resource post user:references
###
FactoryBot.define do
  factory :user do
    title { "MyString" }
    user
  end
end
```


`attributes_for(:post)` will generate:
```
{ :title => "MyString" }
```
`build(:post).attributes` will generate:
```
{ :title => "MyString", :user_id => 1 }
```

I think we would want the latter when testing the creation of a resource. Without supplying the user_id reference, we'll receive this error when trying to save the resource:
```
  1) posts#create basic create works
     Failure/Error: expect(response.status).to eq(201), response.body
       {"errors":[{"code":"unprocessable_entity","status":"422","title":"Validation Error","detail":"User must exist","source":{"pointer":"/data/relationships/user"},"meta":{"attribute":"user","message":"must exist","code":"blank"}}]}

  2) PostResource creating works
     Failure/Error: expect(instance.save).to eq(true), instance.errors.full_messages.to_sentence
       User must exist
```

As @wadetandy said, generators exist to help users get up to speed quickly but not solve edge cases. I 100% agree. I also think the above is a common use case. Using attributes_for adds an extra step for the user to create the associated objects and reference the ids as part of the payload. I think that this results in a better/clearer spec. However, I'm also torn because I feel like the specs should pass after the generator is run.

For an example, I've created a bash script to create a sample blog app with the generator both ways: 
https://gist.github.com/sideshowbandana/b4066e8096dd6f45d78becb37a5c4bb1

richmolj:
@sideshowbandana I actually think this has more to do with the convention for writing foreign keys than anything else. Regardless of anything with FactoryBot or generators, I'd argue that:

```ruby
# POST /posts

{
  type: 'posts',
  attributes: { user_id: 123 }
}
```

Is not ideal. Instead, I'd suggest

```ruby
{
  type: 'posts'
  attributes: { ... },
  relationships: {
    user: {
      data: { type: 'users', id: 123 }
    }
  }
}
```

as a better option. This way we don't expose the foreign key (an implementation detail) to the API. In addition, we might have special logic when associating resources that wouldn't elsewise fire.

It's not the biggest deal in the world, and I've certainly exposed the FK as a convenience before. But I don't think it's a best practice we want our generators to promote. In fact, [our sample app will do this instead of exposing the foreign key](https://github.com/graphiti-api/employee_directory/blob/master/spec/resources/position/writes_spec.rb#L15-L29).

In addition, this is a Rails 5 thing as much as anything else. [Rails 5 made belongs_to required by default](https://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html). Many people turn this off in test mode with `Rails.application.config.active_record.belongs_to_required_by_default = false`.

I think we should better document this and the various ways to overcome, but I'm not sure if it's up to the generators. Ultimately your activerecord associations would cause the spec failure here, not anything Graphiti is doing. So I'm OK with a little extra work.

It would be nice to support something like `rails g graphiti:resource Post title:string user:belongs_to`. But I think `attributes_for` is the correct approach as it stands right now.

I'll hold off on merging until I can get your thoughts on this - what do you think?

sideshowbandana:
Fair enough. I say we push this and think about a good way to message this to users in a separate PR.

richmolj:
@wadetandy spoke to @sideshowbandana in Slack and we're good to go here. [We do have some documentation on this issue](https://graphiti-api.github.io/graphiti/guides/concepts/testing#required-belongs-to) though I'd like to raise its profile. This LGTM, but wanted to check with you before merge.

wadetandy (approved):

wadetandy:
This looks good and I agree with follow up actions discussed above.  Thanks @sideshowbandana for the back and forth on this and for your patience while we worked through it!  Solid feature!

PR #53, opened 2019-01-07
Add around_persistence

richmolj:
For when you want to modify attributes but also post-process the saved
record:

```ruby
around_persistence :attach_file

def attach_file(attributes)
  file = attributes.delete(:file)
  model = yield
  attach(model, :files, file) if file
end
```

### DIFF
--- a/lib/graphiti/resource/persistence.rb
+++ b/lib/graphiti/resource/persistence.rb
@@ -44,6 +44,14 @@ module Graphiti
           end
         end
 
+        def around_persistence(method = nil, only: [:create, :update], &blk)
+          if blk
+            raise Errors::AroundCallbackProc.new(self, 'around_persistence')
+          else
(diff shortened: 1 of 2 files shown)


lib/graphiti/resource/persistence.rb
```diff
@@ -44,6 +44,14 @@ def around_save(method = nil, only: [:create, :update], &blk)
           end
         end
 
+        def around_persistence(method = nil, only: [:create, :update], &blk)
+          if blk
+            raise Errors::AroundCallbackProc.new(self, 'around_persistence')
```
wadetandy, line 49: I forget why we have to do this?

wadetandy (approved):
Looks good, just the one question.

Also, thoughts on adding sugar for `{before,after,around}_{create,update}` hooks?:

```ruby
def around_create(method, &blk)
  around_persistence method, only: [:create], &blk
end
```

PR #54, opened 2019-01-08
Print response.body if status code is incorrect

sideshowbandana:
Show the error messages in failed specs when create/update/delete
fails.


richmolj:
What were the scenarios where you found this helpful? I ask because I rarely get these type of failures.

sideshowbandana:
I found it useful for validation or foreign key errors. The way that it's currently written only gave me info that the object was not getting saved but didn't have any information as to why. Failing on the response code and printing the body bubbles that up to the user a without having to debug in.

richmolj:
I'll buy that. Thanks! 👏

PR #60, opened 2019-01-11
Pass in the method to the before_commit hook

tylerkahn:
There might be different conditions under which the client would like to raise an exception depending on the method (e.g. :create vs :destroy).

### DIFF
--- a/lib/graphiti/resource.rb
+++ b/lib/graphiti/resource.rb
@@ -112,7 +112,7 @@ module Graphiti
 
     def before_commit(model, method)
       hook = self.class.config[:before_commit][method]
-      hook.call(model) if hook
+      hook.call(model, method) if hook
     end
 
     def transaction
--- a/spec/integration/rails/before_commit_spec.rb
+++ b/spec/integration/rails/before_commit_spec.rb
@@ -62,10 +62,12 @@ if ENV["APPRAISAL_INITIALIZED"]
 
         attribute :first_name, :string
 
-        before_commit only: [:create] do |employee|
-          Callbacks.add(:create, employee)
-          if $raise_on_before_commit[:employee]
-            raise 'rollitback'
+        before_commit only: [:create] do |employee, method|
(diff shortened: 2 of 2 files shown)


wadetandy:
Seems like you could achieve the same by having two hooks, with each using an “only” argument?

tylerkahn:
I hadn't considered that prior to modifying the spec to be honest. IMO it's not clear from the documentation that you can 'stack' them in that manner i.e.

    before_commit only: [:create] do |resource|
    end
    
    before_commit only: [:update] do |resource|
    end

I had assumed that declaring before_commit multiple times would keep overriding the previous declarations. My understanding (maybe wrong) is that it **will** actually do that kind of overriding if you do this:

    before_commit only: [:create] do |resource|
      puts "First"
      resource
    end
    
    before_commit only: [:create] do |resource|
      puts "Second"
      resource
    end

As I understand it, this will print out only "Second" and not "First\nSecond".

IMO the behavior I'm adding here makes it less ambiguous (also it lends itself nicely to a bit cleaner of an abstraction). I'm using the before_commit hook to do authorization with CanCan so my ApplicationResource now has this in it:

    before_commit do |resource, method|
      authorize! method, resource
    end

richmolj:
@tylerkahn I support this, but think it is part of a slightly larger bit of work. Instead of yielding `method`, I'd like to yield `metadata` - and you'd access `metadata.method`.

We have a few other use cases where this might be helpful, like `metadata.caller_model`. This way we can hang bits of information on the same yielded object, adding to it as we find use cases. I'd like this to be consistent for *all* hooks, not just `before_commit`.

Separately, the hooks need A) documentation and B) *should* be stackable in the way you describe (let's fix it if they are not). I can take care of A this weekend, but any help on `metadata` or hook stacking would be appreciated ❤️

richmolj:
Addressed in https://github.com/graphiti-api/graphiti/pull/87

PR #61, opened 2019-01-17
Add Remote Resources

richmolj:
If you run more than 1 Graphiti API, eventually you need to make
cross-api calls. A good example is a Blog API and a User API - the blog
may want to sideload `author` via the Users API. Breaking apps apart
like this leads to conflict-free deployments and speedy test suites.

You can now do this:

```ruby
belongs_to :author, remote: 'http://my-api.com/api/v1/users'
```

Which is equivalent to:

```ruby
belongs_to :author

class AuthorResource < ApplicationResource
  self.remote = 'http://my-api.com/api/v1/users'
end
```

Everything works as normal, **including Deep Queries**. Note remote
resources only support read operations.

One special case is additional sideloads. Let's say we *also* want to
get that user's Account. Should we load Account locally, via the Blog
API, or remotely, via User API?

If you explicitly define an association, it will be treated as local. If
you do not, it will be treated as remote:

```ruby
class AuthorResource < ApplicationResource
  # fetched locally
  has_many :blogs

  # /blogs?include=author.account
  # here 'account' is loaded remotely in the User API
end
```

If the remote API gives an error, we'll re-raise it in the local API. To
get more robust errors, [turn on raw errors](https://graphiti-api.github.io/graphiti/guides/concepts/error-handling#displaying-raw-errors)
when the request comes from a remote API.

By default, we'll forward all HTTP headers to the remote API. You can
customize your own headers:

```ruby
 # app/resources/author_resource.rb
def request_headers
  { Authorization: 'Bearer "abc123"' }
end
```

You can also customize timeouts:

```ruby
 # app/resources/author_resource.rb
self.open_timeout = 10
self.timeout = 20
```

We use [Faraday](https://github.com/lostisland/faraday) to make the
request. **You must add faraday to your Gemfile to use this feature**.

To customize the faraday request:

```ruby
 # app/resources/author_resource.rb
 def make_request(url)
   super do |req|
     req.options.foo = 'bar'
   end
 end
```

Remote resources do not currently support links or endpoint validation.


lib/graphiti/resource/remote.rb
```diff
@@ -0,0 +1,48 @@
...
+          context.request.headers.to_h.reject { |k, v| k.include?('.') }
+        else
+          {}
+        end
+      end
+
+      def make_request(url)
+        headers = request_headers.dup
+        headers['Content-Type'] = 'application/vnd.api+json'
+        Faraday.get(url, nil, headers) do |req|
```
wadetandy: Since we're making this an optional dependency, would be nice to require and raise:

```ruby
unless defined?(Faraday)
  raise "Faraday not defined. Please require the 'faraday' gem to use remote resources"
end
```

lib/graphiti/resource/remote.rb
```diff
@@ -0,0 +1,48 @@
...
+          yield req if block_given? # for super do ... end
+          req.options.timeout = timeout if timeout
```
wadetandy, line 42: Instead of adding a bunch of faraday-specific class attributes and moving them over (and dealing with the PRs that will come in asking for new ones), I'd rather see something like `def faraday_request_options(options)` which lets users modify whatever they need.

lib/graphiti/sideload.rb
```diff
@@ -70,10 +75,25 @@ def self.link(&blk)
       self.link_proc = blk
     end
 
+    # Todo ApplicationResource
+    def create_remote_resource
```
wadetandy, line 80: Should we raise an error or maybe a warning if:

```
class PostResource
  self.remote = 'http://my-api.com/api/v1/post'
end

class UserResource
  has_many :posts, remote: 'http://my-api.com/api/v1/post'
end
```

As the user probably meant to use the existing Post resource.  Also we shouldn't allow a user to specify both `remote` and `resource` to the same sideload.

lib/graphiti/adapters/jsonapi.rb
```diff
@@ -0,0 +1,89 @@
+module Graphiti
+  module Adapters
+    class JSONAPI < ::Graphiti::Adapters::Null
```
wadetandy: This is probably more specifically a `GraphitiAPI` resource, as it assumes our request/response features, yes?

lib/graphiti/errors.rb
```diff
@@ -29,6 +45,18 @@ def message
       end
     end
 
+    class RemoteWrite < Base
+      def initialize(resource_class)
+        @resource_class = resource_class
+      end
+
+      def message
```
wadetandy, line 67: 👍

wadetandy (commented):
Have to run and will follow up with more later, but a few comments. This is great though and I'm super excited for this!

lib/graphiti/resource/remote.rb
richmolj: Good call, force-pushed

lib/graphiti/adapters/jsonapi.rb
richmolj: Agreed, force-pushed

lib/graphiti/resource/remote.rb
richmolj: I actually see `timeout` and `open_timeout` as not specific to Faraday. The underlying `Net::HTTP` implementation actually uses these values, so I view it as a general purpose thing.

Since faraday is geared more towards middleware, I think customizing faraday with a block instead of a hash is more appropriate:

```ruby
def make_request(url)
  super do |req|
    # customize faraday request
  end
end
```

lib/graphiti/adapters/graphiti_api.rb
```diff
@@ -0,0 +1,89 @@
...
+            { message: raw['message'], backtrace: raw['backtrace'] }
+          else
+            { message: "#{error['title']} - #{error['detail']}" }
+          end
+        end.compact
+        raise Errors::Remote.new(url, errors)
+      end
+
+      def build_url(scope)
+        url    = resource.remote
```
wadetandy: Thinking about a more robust `UserResource < UsersApiResource` subclassing setup for multiple resources from the same API, I'd like to be able to set either `remote` to be very explicit or something like `remote_base_url` on a base class, which would allow each child resource to either infer or explicitly set its resource name to append to build `remote`.

spec/remote_resource_spec.rb
```diff
@@ -0,0 +1,511 @@
...
+            has_many :positions, resource: PORO::PositionResource
+          end
+        end
+
+        it 'does not pass to the remote api' do
+          assert_params('include=teams')
+        end
+      end
+
+      context 'and the association is NOT defined locally' do
```
wadetandy, line 340: Thought of one more branch in this condition.  If I have a remote resource, I can define its attributes, serializers, sideloads, etc, but only for the "root" remote resource.  If I explicitly define a sideload and that resource is also a remote resource for the same API, it will do two requests, yes?  Ideally we'd be able to make this do a single request but still leverage the local resource definition logic for serialization and such.

lib/graphiti/resource/remote.rb
wadetandy: Ahh I had actually forgotten `open_timeout` worked on underlying `Net::HTTP` as well. I'm fine with this then if we aren't planning on extending this option set any further.

lib/graphiti/sideload.rb
richmolj: I added a check that we can't pass both `resource` and `remote` 👍 

That said, I think the code snippet you posted is WAD and we should solve that with documentation. If someone is surprised by this behavior, it means they don't understand `remote` to begin with.

lib/graphiti/adapters/graphiti_api.rb
richmolj: Added `remote_base_url`, which is *optional*. `remote` can be a path or a full url depending on your use of this.

spec/remote_resource_spec.rb
richmolj: I added a test for this. The way it works now is if the association is defined locally, but the associated is a remote resource with the same `remote_base_url`, it will be one request to the remote instead of two.

wadetandy:
Ok, I'm happy with all this. Great feature.

Issue #62, opened 2019-01-18
Resolve polymorphic resource before typecasting the attributes in Util::Persistence initializer


mihaimuntenas:
Using polymorphic resources, the expectation one has is that the child resources can define additional attributes and relations (as per tasks in employee_directory example), whilst still using the base resource to create them - using appropriate payload.

Unfortunately this will not work for attributes as the Util::Persistence in the initializer typecasts the attributes  before resolving the `@resource` to the child resource; in which case the typecast fails because the attribute is only defined by the child resource.

Should resolving the `@resource` happen before the attributes typecast?
Code from line 19->24 be before `@attributes.each_pair` at line 15?

https://github.com/graphiti-api/graphiti/blob/ae91899884f0a0e2a42d0290e8c5d8f3a7dcc343/lib/graphiti/util/persistence.rb#L19

mihaimuntenas:
PR: https://github.com/graphiti-api/graphiti/pull/63

mihaimuntenas:
PR merged; closing.

PR #63, opened 2019-01-18
Fix Issue #62

mihaimuntenas:
Resolve polymorphic child resource before typecasting attributes.

### DIFF
--- a/lib/graphiti/util/persistence.rb
+++ b/lib/graphiti/util/persistence.rb
@@ -12,15 +12,15 @@ class Graphiti::Util::Persistence
     @relationships = relationships
     @caller_model  = caller_model
 
-    @attributes.each_pair do |key, value|
-      @attributes[key] = @resource.typecast(key, value, :writable)
-    end
-
     # Find the correct child resource for a given jsonapi type
     if meta_type = @meta[:type].try(:to_sym)
       if @resource.type != meta_type && @resource.polymorphic?
         @resource = @resource.class.resource_for_type(meta_type).new
       end
+    end
+
+    @attributes.each_pair do |key, value|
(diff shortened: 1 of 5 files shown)


richmolj:
Looks great, thanks for this PR ❤️ Do you think you could add a test?

richmolj:
I'm actually trying to leave the Util classes more or less private, and instead test the real-world functionality. I think we could do this in [polymorphic_resource_spec.rb](https://github.com/graphiti-api/graphiti/blob/master/spec/polymorphic_resource_spec.rb) or [polymorphism_spec.rb](https://github.com/graphiti-api/graphiti/blob/master/spec/polymorphism_spec.rb), what do you think?

mihaimuntenas:
Well, my issue showed up only when the Persistence instance was created, before it was run, from within the Resource ... 
I also checked a bit the Resource.build, to build a derived resource through the base one, but the resulting Proxy, had the Resource of the base type, not of the derived type (which is somehow in line with ActiveRecord). In the end, the not-persisted object of the base resource can be transformed to a different derived one (eg: initialize a CreditCard with Visa attributes, you should still be able to change it to a MasterCard - the actual "locking" of the type should be done upon persistence).
Given this, that the actual polymorphic class is determined within the Persistence and not upon building the Resource, made me implement the spec on Persistence, rather than Resource.

mihaimuntenas:
I also added some specs for the public API.
Though, even if the Persistance class should be private API, I would still keep the specs for that - as the actual determination is done there and the issue (sequence of operations) was within the initializer of the Persistence class.

richmolj:
Works for me, thanks a bunch! ❤️

PR #66, opened 2019-01-19
Assign correctly on string/integer mismatch

richmolj:
Occasionally there's the scenario where the PK and FK have a
string/integer mismatch. The most common is with Remote Resources,
where the local API has an integer FK but the remote API has a string PK
(because all ids in JSONAPI are strings).

### DIFF
--- a/lib/graphiti/sideload/belongs_to.rb
+++ b/lib/graphiti/sideload/belongs_to.rb
@@ -48,6 +48,16 @@ class Graphiti::Sideload::BelongsTo < Graphiti::Sideload
   end
 
   def children_for(parent, map)
-    map[parent.send(foreign_key)]
+    fk = parent.send(foreign_key)
+    children = map[fk]
+    return children if children
+
+    keys = map.keys
+    if fk.is_a?(String) && keys[0].is_a?(Integer)
+      fk = fk.to_i
+    elsif fk.is_a?(Integer) && keys[0].is_a?(String)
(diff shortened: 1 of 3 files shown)


richmolj:
@wadetandy could you take a look at this one?

lib/graphiti/sideload/belongs_to.rb
```diff
@@ -48,6 +48,16 @@ def child_map(children)
   end
 
   def children_for(parent, map)
-    map[parent.send(foreign_key)]
+    fk = parent.send(foreign_key)
```
wadetandy, line 51: This is complex enough logic shared between the two locations (with pk vs fk) that it would be nice to extract to util method/class

spec/sideload_spec.rb
```diff
@@ -330,6 +330,37 @@ class SideloadSpecEmployee
...
+            PORO::Employee.new(id: '2')
+          ]
+        end
+
+        it 'still works' do
+          instance.assign(employees, positions)
+          expect(employees[0].positions).to eq(positions[0..1])
+          expect(employees[1].positions).to eq([positions.last])
+        end
+      end
```
wadetandy, line 363: Would add a check for null keys not matching when they shouldn’t. Think it works here anyway but will prevent regressions

wadetandy (approved):
Couple comments but looks good overall.

PR #68, opened 2019-01-19
Treat empty strings for expected integers as nil

elDub:
Allow nullification of integer attributes when empty string is passed.

### DIFF
--- a/lib/graphiti/types.rb
+++ b/lib/graphiti/types.rb
@@ -57,23 +57,25 @@ module Graphiti
     end
 
     Integer = create(::Integer) do |input|
-      Dry::Types['coercible.integer'][input] if input
+      Dry::Types['coercible.integer'][input] unless input.blank?
     end
 
     # The Float() check here is to ensure we have a number
     # Otherwise BigDecimal('foo') *will return a decimal*
     ParamDecimal = create(::BigDecimal) do |input|
-      Float(input)
-      input = Dry::Types['coercible.decimal'][input]
-      Dry::Types['strict.decimal'][input]
+      unless input.blank?
+        Float(input)
+        input = Dry::Types['coercible.decimal'][input]
+        Dry::Types['strict.decimal'][input]
+      end
     end
 
     PresentInteger = create(::Integer) do |input|
-      Dry::Types['coercible.integer'][input]
+      Dry::Types['coercible.integer'][input] unless input.blank?
     end
 
     Float = create(::Float) do |input|
-      Dry::Types['coercible.float'][input] if input
+      Dry::Types['coercible.float'][input] unless input.blank?
     end
 
     PresentParamsHash = create(::Hash) do |input|
--- a/spec/persistence_spec.rb
+++ b/spec/persistence_spec.rb
@@ -983,16 +983,14 @@ RSpec.describe 'persistence' do
         expect(save(nil)).to eq(nil)
       end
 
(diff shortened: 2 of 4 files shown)


richmolj:
This looks great @elDub, thanks ❤️ ! I think the only thing this needs is to cover the other types as well?

elDub:
I added similar code to other numerics and added/updated tests.

richmolj:
I think the biggest consideration is performance because we'd be calling `#blank?` a bunch more times and most people aren't using [fast_blank?](https://github.com/SamSaffron/fast_blank).

Another scenario: it makes sense that *filtering* wants this behavior, but I could see serializing `null` instead of `""` to be a surprising result.

@wadetandy thoughts?

wadetandy:
Is the `blank?` check actually the right one?:

```ruby
"".blank? # => true
"".empty? # => true
"    ".blank? # => true
"    ".empty? # => false
```

Not only is the `empty?` check _much_ faster, but it's more correct for the intended use-case.

wadetandy:
Stepping back from a second from the specific implementation details, I'd like to ask a bit about this from an API usability perspective.  Curious as to what pain you find with the current way of passing either `null` or omitting a key entirely to null a particular number field?  Is it providing some unexpected behavior?  I see this path also potentially providing unexpected behavior so want to understand the problem a bit better.

elDub:
Then is this a problem with Spraypaint? When I clear out a numeric value in the client using Spraypaint, it passes an empty string which is failing in Graphiti when the attribute is defined as `:integer`.

wadetandy:
When you say "clearing out", how do you mean? Are you binding the field directly to a text input, or are you setting it to `null` directly in javascript?  The latter case behaves as expected in the test suite, but the former may need an enhancement.

elDub:
I am binding the field to a text input in VueJS.

elDub:
So your change will make it send this?
```attributes: {numberfield: null}```

Instead of this?
```attributes: {numberfield: ""}```

Does Spraypaint have a way to define what type an attribute is, or will this change apply to _all types_, meaning if someone clears out a regular string value in the UI that it too will receive `null`?

**Comment by wadetandy** [comment]
Per [this PR](https://github.com/graphiti-api/spraypaint.js/pull/12), we will actually listen when you specify the type of the attribute:

```javascript
class Person extends ApplicationRecord {
  @Attr firstName
  @Attr({ type: Number }) age
}
```

If a field of type `Number` is an empty string, it will send `null` to the server on persist.

Note that this does not handle the filtering use-case.  `Person.where({ 'age': null })` will currently drop the filter altogether.

elDub:
Cool. I was not aware of the `type:` option for an attribute. Should this also be available for date objects?

wadetandy:
@elDub I think in the medium term, yes, but those are a bit more of an issue since there's both serialization and deserialization casting to handle (the wire sees them as just strings).  I'd rather not tackle that the way this part of the codebase currently works, and I've got a general (de)serialization cleanup/rework I'm planning to handle in the near future.

wadetandy:
@elDub Just pushed [spraypaint 0.5.0](https://github.com/graphiti-api/spraypaint.js/releases/tag/v0.5.0), which has the fix for this issue.  If this fixes your issue then I think we can close this PR.  You may still need to manually null out your empty string dates for the time being though.

wadetandy:
Closing, as we've added a fix to spraypaint for this behavior.

PR #72, opened 2019-01-22
Links: Failing test for empty belongs_to

ziggurat:


richmolj:
Looks great, thanks! Minor, but could you make this `.update_attributes` instead of `.create`? https://github.com/graphiti-api/graphiti/blob/acb06fda1aa34a96d2ca86b0a141fd2d45444da4/spec/serialization_spec.rb#L1224

The reasoning is this way we're only updating the bit of information relevant to the test, instead of redefining the whole object and possibly getting an unexpected side effect

ziggurat:
Yes! My bad! Will change it as soon as I can.

ziggurat:
Changed!

richmolj:
Fixed in master!

PR #73, opened 2019-01-23
Allow polymorphic_belongs_to to ignore possible types

Dishwasha:
By allowing possible types to be ignored, multiple polymorphic_belongs_to
relationships can be built from the same foreign key as well as just
ignoring certain types stored in the grouping field's data in general.

This allows for resource definitions with polymorphic_belongs_to definitions like:
```
# Call <employee resource uri>?include=credit_card,payment_processor

  polymorphic_belongs_to :credit_card do
    group_by(:credit_card_type, exclude: [:Paypal]) do
      on(:Visa)
      on(:Mastercard)
    end
  end

  polymorphic_belongs_to :payment_processor do
    group_by(:credit_card_type, only: [:Paypal]) do
      on(:Paypal).belongs_to :payment_processor,
        resource: PaypalResource,
        foreign_key: :credit_card_id
    end
  end
end
```

### DIFF
--- a/lib/graphiti/sideload/polymorphic_belongs_to.rb
+++ b/lib/graphiti/sideload/polymorphic_belongs_to.rb
@@ -15,9 +15,24 @@ class Graphiti::Sideload::PolymorphicBelongsTo < Graphiti::Sideload::BelongsTo
   class Grouper
     attr_reader :field_name
 
-    def initialize(field_name)
+    def initialize(field_name, opts={})
       @field_name = field_name
       @groups = []
+      @except = Array(opts[:except]).map(&:to_sym)
+      @only = Array(opts[:only]).map(&:to_sym)
+    end
+
+    def expected?(group_sym)
+      @only.empty? || @only.include?(group_sym)
+    end
+
+    def excluded?(group_sym)
+      @except.include?(group_sym)
+    end
+
+    def ignore?(group_name)
+      group_sym = group_name.to_sym
+      !expected?(group_sym) || excluded?(group_sym)
     end
 
     def on(name, &blk)
@@ -61,8 +76,8 @@ class Graphiti::Sideload::PolymorphicBelongsTo < Graphiti::Sideload::BelongsTo
     :"#{name}_id"
   end
 
-  def self.group_by(name, &blk)
-    self.grouper = Grouper.new(name)
+  def self.group_by(name, opts={}, &blk)
+    self.grouper = Grouper.new(name, opts)
     self.grouper.instance_eval(&blk)
   end
 
@@ -80,7 +95,7 @@ class Graphiti::Sideload::PolymorphicBelongsTo < Graphiti::Sideload::BelongsTo
 
   def resolve(parents, query, graph_parent)
     parents.group_by(&grouper.field_name).each_pair do |group_name, group|
-      next if group_name.nil?
+      next if group_name.nil? || grouper.ignore?(group_name)
 
       match = ->(c) { c.group_name == group_name.to_sym }
       if sideload = children.values.find(&match)
--- a/spec/sideloading_spec.rb
+++ b/spec/sideloading_spec.rb
@@ -337,9 +337,14 @@ RSpec.describe 'sideloading' do
   describe 'polymorphic_belongs_to macro' do
     let!(:visa) { PORO::Visa.create }
     let!(:mastercard) { PORO::Mastercard.create }
+    let!(:paypal) { PORO::Paypal.create }
     let!(:employee2) do
       PORO::Employee.create credit_card_type: 'Mastercard',
           credit_card_id: mastercard.id
+    end
+    let!(:employee3) do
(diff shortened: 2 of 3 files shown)


Dishwasha:
@richmolj @wadetandy Please let me know what you think of the interface and test coverage.

lib/graphiti/sideload/polymorphic_belongs_to.rb
```diff
@@ -15,9 +15,14 @@ def method_missing(name, *args, &blk)
   class Grouper
     attr_reader :field_name
 
-    def initialize(field_name)
+    def initialize(field_name, opts={})
       @field_name = field_name
       @groups = []
+      @ignore_groups = Array(opts[:ignore_groups])
```
richmolj: This looks implemented correctly, but I'd prefer `ignore` over `ignore_groups`

spec/sideloading_spec.rb
```diff
@@ -337,10 +337,29 @@ def self.name
...
+    let!(:paypal) { PORO::Paypal.create }
     let!(:employee2) do
       PORO::Employee.create credit_card_type: 'Mastercard',
           credit_card_id: mastercard.id
     end
+    let!(:employee3) do
+      PORO::Employee.create credit_card_type: 'Paypal',
+                            credit_card_id: paypal.id
+    end
+    let(:assert_correct_response) do
```
richmolj: Typically use `let` for setup and `def assert_correct_response` for assertions

spec/sideloading_spec.rb
```diff
@@ -337,10 +337,29 @@ def self.name
...
+      expect(credit_card(0)).to eq({
+                                     'type' => 'visas', 'id' => '1'
+                                   })
+      expect(credit_card(1)).to eq({
+                                     'type' => 'mastercards', 'id' => '1'
+                                   })
+      expect(included[0].jsonapi_type).to eq('visas')
+      expect(included[0].relationships).to eq({
+                                                'visa_rewards' => { 'meta' => { 'included' => false } }
+                                              })
```
richmolj: I know it's minor but I'd prefer these not indent

spec/sideloading_spec.rb
```diff
@@ -405,6 +413,59 @@ def assert_correct_response
...
+
+      it 'does not create or require ignored types' do
+        sl = resource.sideload(:credit_card)
+        expect(sl.resource).to be_polymorphic
+        expect(sl.resource.class).to be_abstract_class
+        expect(sl.children.keys).not_to include(:paypal)
+      end
+    end
+
+    context 'when multiple macros defined on the same keys' do
```
richmolj, line 457: I do like the test on 417, but would like a more integration-y test like this one as well (actually querying and asserting the json doesn't include the relationship, but without the additional setup needed for this multi-macro use case)

richmolj (commented):
Looks great! My comments are mostly stylistic, but if you bear with me I think we can get this merged shortly.

One thought is to consider `only/except` keywords instead of `ignore`. I'm find with `ignore` since it's a little more expressive, but `only/except` don't lose much and are more widely seen elsewhere.

lib/graphiti/sideload/polymorphic_belongs_to.rb
wadetandy: Tend to agree with the above comment around `except` instead of `ignore`, since that matches our options elsewhere in the library/rails defaults.

lib/graphiti/sideload/polymorphic_belongs_to.rb
```diff
@@ -15,9 +15,14 @@ def method_missing(name, *args, &blk)
...
+    end
+
+    def ignore_group?(group_name)
+      @ignore_groups.include?(group_name.to_sym)
```
wadetandy: If casting the group name `to_sym` on check, I would do the same to the ignore list when assigning in the initializer.

spec/sideloading_spec.rb
Dishwasha: fixed.  Couldn't remember if rspec def defined outside a context and redefined inside a nested context would be fragile depending on execution order.

spec/sideloading_spec.rb
Dishwasha: thought I caught them all....

lib/graphiti/sideload/polymorphic_belongs_to.rb
Dishwasha: Changed.  Please let me know what you think.

spec/sideloading_spec.rb
Dishwasha: I added two more specs; not sure if this is what you were looking for.

lib/graphiti/sideload/polymorphic_belongs_to.rb
Dishwasha: done

richmolj (approved):
Awesome 🎉 ! Thanks so much for this ❤️ !

PR #78, opened 2019-01-30
Fix polymorphic remote queries

richmolj:
There were two issues here.

The first was that the Query object references the top-level of the
polymorphic sideload (e.g. `credit_card`), and not the child type
implementation (e.g. `visa`). This is problematic when deriving the
query chain that we need to pass to the remote API. Solution is to move
the chain logic out of Query and into `Util::RemoteParams`, which is
where it was actually used and a better place for the code either way.

The second was with polymorphic relationships modifying the params. When
you modify params, we mutate `query.hash` - not a problem and even
desirable/correct for non-polymorphic associations. But when
polymorphic, that hash is going to be passed again to each sub-relation
(`mastercard`, `visa`, etc.). So mastercard modifies params, and visa
incorrectly gets that modification. Solution is to refresh `query.hash`
when processing each level of the polymorphic relation.

### DIFF
--- a/lib/graphiti/query.rb
+++ b/lib/graphiti/query.rb
@@ -210,24 +210,6 @@ module Graphiti
       not [false, 'false'].include?(@params[:paginate])
     end
 
-    # If this is a remote call, we don't care about local parents
-    def chain
-      if @resource.remote
-        top_remote_parent = parents.find { |p| p.resource.remote? }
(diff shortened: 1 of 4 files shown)


richmolj:
@wadetandy

wadetandy (approved):

PR #79, opened 2019-01-31
Fix issue with multiple sidepost destroys

richmolj:
When validating, we would iterate over the relationships and compare them to their spot
in the payload. This normally works because we iterate both the same
way. However, we never associate objects that have been destroyed or
disassociated - which means the comparison will be incorrect and even
blow up when multiple objects are being persisted.

The solution is to skip these guys when iterating. This means we don't
recursively validate a destroyed object's associations, so I'd like us to
move to a pattern closer to what's happening in `before_commit`. But
this solves the problem for now and makes anything else an extreme edge
case.

Addresses https://github.com/graphiti-api/graphiti/pull/77/files


richmolj:
@wadetandy @pariv

PR #80, opened 2019-01-31
Render nil link when belongs_to FK is nil

richmolj:
The jsonapi-rb library is always going to render this section of the
payload (when applicable), but the actual URL should be `nil` when a
belongs_to relationship has a `nil` foreign key. This probably makes the
most sense, so clients can still parse the payload as normal, but we
don't render an invalid url.


richmolj:
Addresses https://github.com/graphiti-api/graphiti/pull/72

Issue #82, opened 2019-01-31
Pagination Links Error


richmolj:
@sideshowbandana one of my users had an issue with this right after release.

```
ERROR: NoMethodError: undefined method `/' for nil:NilClass
graphiti (1.0.rc.10) lib/graphiti/delegates/pagination.rb:40:in `last_page'
graphiti (1.0.rc.10) lib/graphiti/delegates/pagination.rb:15:in `block in links'
graphiti (1.0.rc.10) lib/graphiti/delegates/pagination.rb:13:in `tap'
graphiti (1.0.rc.10) lib/graphiti/delegates/pagination.rb:13:in `links'
graphiti (1.0.rc.10) lib/graphiti/renderer.rb:47:in `block in render'
graphiti (1.0.rc.10) lib/graphiti.rb:141:in `block in broadcast'
```

The larger issue is that we should have these links off by default. I did that [here](https://github.com/graphiti-api/graphiti/pull/81).

There also seemed to be some spec code that was problematic locally https://github.com/graphiti-api/graphiti/blob/47cea9c19731ae84ccf6112d66c54c69b5e67c41/spec/spec_helper.rb#L70-L73 

the merged code was `rspec` on line 70 and both seem to be in the wrong spot.

Could you look into these issues?

PR #83, opened 2019-02-01
Handle invalid item_counts for #last_page method

sideshowbandana:
- Ensure that #item_count always returns an integer.
- Add specs for page_size and item_count being set to 0
- Add spec for item_count when resource.stat returns nil

FIXES: https://github.com/graphiti-api/graphiti/issues/82


sideshowbandana:
@richmolj I took a stab at addressing #82. It's a bit odd to me how it occurred though...

https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/delegates/pagination.rb#L49-L58 should prevent `@item_count` from ever being nil unless for some reason `nil.kind_of?(Numeric)` is returning true?

I've also removed the spec_helper changes. Those weren't supposed to be submitted as part of the PR. However, I am curious what pattern you would like me to follow for shared_contexts. Currently I'm using a require_relative https://github.com/graphiti-api/graphiti/blob/master/spec/delegates/pagination_spec.rb#L2. It looks like the other shared_contexts are brought in via `graphiti_spec_helpers` Should I make a PR that moves that code ? What's the process for making multiple dependent changes?

richmolj:
> should prevent @item_count from ever being nil 

I believe this is because `@item_count` instance variable is the correct value, but `item_count` is not. The method isn't returning `@item_count`, it's returning the result of the `unless` block. We should probably try to remove some of the spec stubbing to expose this issue.

For shared context, does `spec/support/pagination_link_helper.rb` work?

sideshowbandana:
> it's returning the result of the `unless` block.

Ah. That's it. I thought I had a return statement at the end.  >_<

Updated the patch and renamed the spec helper

richmolj (approved):
Thanks for addressing this!

PR #84, opened 2019-02-01
Formalize persistence overrides

richmolj:
* Added ability to override adapter destroy logic
* Additional tests for existing overrides


richmolj:
@wadetandy

PR #85, opened 2019-02-01
Allow multiple create/update before_commit hooks

richmolj:


richmolj:
@wadetandy

wadetandy:
Looks good as an improvement on the original implementation, though considering all other hooks are using a different code path, I'd like to move over to that sooner rather than later.

wadetandy (approved):

richmolj:
>  I'd like to move over to that sooner rather than later.

There's probably *something* we can do, but it isn't as simple as just using the same codepath. The attribute/save hooks on the resource are when we're performing an operation within the resource. The before_commit actually happens outside of the resource - we have to process the full graph before this can be called. That's ultimately why it will always be a separate codepath on some level.

PR #87, opened 2019-02-02
Optionally yield metadata on persistence hooks

richmolj:
Now, all persistence hooks will optionally yield metadata - a hash
containing things like the method, temp id, further relationships,
the calling model in the graph. We can continue hanging things on
metadata as we get use cases.

The main use case is referencing the "caller model" - the parent in the
graph. If we're saving an Employee and her Positions in a single
request:

```ruby
class PositionResource < ApplicationResource
  before_commit do |position, meta|
    meta[:caller_model] # => #<Employee>
  end
end
```

This is also true for override methods `#build`, `#assign_attributes`,
`#save`, and `#destroy` - all now accept `meta` as an optional argument,
so it is accessible when overriding.

A notable case is `destroy`. We do not currently handle nested destroys,
and the code path for destroying a top-level entity is different than
when sideposting a deletion. We accomodate both use cases here, but
notable.

### DIFF
--- a/lib/graphiti/resource/persistence.rb
+++ b/lib/graphiti/resource/persistence.rb
@@ -69,64 +69,64 @@ module Graphiti
         end
       end
 
-      def create(create_params)
+      def create(create_params, meta = nil)
         model_instance = nil
 
-        run_callbacks :persistence, :create, create_params do
-          run_callbacks :attributes, :create, create_params do |params|
-            model_instance = build(model)
-            assign_attributes(model_instance, params)
+        run_callbacks :persistence, :create, create_params, meta do
+          run_callbacks :attributes, :create, create_params, meta do |params|
+            model_instance = call_with_meta(:build, model, meta)
+            call_with_meta(:assign_attributes, model_instance, params, meta)
             model_instance
           end
 
-          run_callbacks :save, :create, model_instance do
-            model_instance = save(model_instance)
+          run_callbacks :save, :create, model_instance, meta do
+            model_instance = call_with_meta(:save, model_instance, meta)
           end
 
           model_instance
         end
       end
 
-      def update(update_params)
+      def update(update_params, meta = nil)
         model_instance = nil
         id = update_params.delete(:id)
 
-        run_callbacks :persistence, :update, update_params do
-          run_callbacks :attributes, :update, update_params do |params|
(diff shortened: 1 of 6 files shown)


richmolj:
@wadetandy please take a look

Addresses https://github.com/graphiti-api/graphiti/pull/60

lib/graphiti/resource/persistence.rb
```diff
@@ -69,74 +69,82 @@ def add_callback(kind, lifecycle, method = nil, only, &blk)
...
       end
 
-      def delete(model_instance)
+      def delete(model_instance, meta = nil)
         adapter.destroy(model_instance)
       end
 
       private
 
+      def call_method(meth, *args)
```
wadetandy: Assuming this is for backwards-compatibility purposes?  Since we haven't released 1.0 yet I'd be in favor of making this a breaking change to avoid a long term support issue. At the very least we could add a deprecation warning if it's the legacy arity, saying that it will go away and give a bit to upgrade.

Additionally, I'd make this method name more descriptive of the purpose since it's specific to handling the `meta` argument.

lib/graphiti/resource/persistence.rb
```diff
@@ -177,16 +191,18 @@ def around_callback_proc(callbacks, index, *args, &blk)
         if callbacks[index + 1]
           proc do
             r = nil
+            args = args[0..0] if method(method_name).arity == 1
```
wadetandy, line 188: Would like to move this arity check + args transformation into a utility method so that it's not repeated so much (and is more descriptive of the purpose).

wadetandy (commented):
The feature addition looks good, but I've got a couple comments about arity wrangling.

lib/graphiti/resource/persistence.rb
richmolj: It's more because >80% of the time you don't care about `meta`. IOW, you should not even have to know about `meta` and do:

```ruby
def save(model)
  Rails.logger.info "saving!"
  super
end
```

And not blow up on accepting one argument instead of two.

Agree on descriptive name though.

lib/graphiti/resource/persistence.rb
richmolj: OK - will do what I can

PR #88, opened 2019-02-02
Support aliases for advanced filtering

richmolj:
Using `alias_attribute` in an ActiveRecord model would work for simpler
queries, but not for queries like `gt` where we dropped down to the
lower-level ARel DSL. This adds support for aliases to all
out-of-the-box filters.


richmolj:
@wadetandy fyi

PR #94, opened 2019-02-20
Allow sideposting unwritable foreign keys

richmolj:
We want to allow sideposting/associating entities, but this should not
require exposing the foreign key as a writable attribute. Instead, we
infer the foreign key from configuration instead of the request, and
make sure this doesn't go through the normal writability check. If the
developer wants to prohibit writing the association, they can do
`has_many :positions, writable: false`.


richmolj:
@wadetandy please take a look

PR #95, opened 2019-02-20
Add after_commit hooks to resources

wadetandy:
Allows users to run logic after database transactions have been
committed. Follows the same logic as before_commit hooks, allowing hooks
to be restricted to specific actions & to run multiple hooks for the
same resource on the same action.


lib/graphiti/resource_proxy.rb
```diff
@@ -154,7 +154,7 @@ def debug_requested?
     private
 
     def persist
-      @resource.transaction do
+      result, after_hooks = @resource.transaction do
```
richmolj: This is kinda odd, and took me a second to understand the different interfaces of `Hooks.run` vs returning`after_hooks`. That said, I get it now and think it's probably the best we can do. A comment here or at the top of `hooks.rb` would be helpful six months from now, I think.

Ideally, we move validations to this same "record a proc for later playback" pattern, which I think will simplify things and cause us to make `Util::Hooks` a more first-class object. IOW I think a private method was/is fine, but this is now smelling like the need for a separate class holding this logic. Not necessarily required for this commit but the direction I'd like to go down.

spec/integration/rails/commit_hooks_spec.rb
```diff
@@ -0,0 +1,406 @@
...
+      {
+        data: {
+          type: 'employees',
+          attributes: { first_name: 'Jane' }
+        }
+      }
+    end
+
+    context 'on create' do
+      it 'fires after validations but before ending the transaction' do
```
richmolj: I like the spec consolidation, but think we need to update some of these strings or add an outer context specifying before/after

spec/integration/rails/commit_hooks_spec.rb
```diff
@@ -0,0 +1,406 @@
...
+      context 'when validation fails' do
+        let(:validation_response) do
+          double('Graphiti::Util::ValidationResponse').as_null_object
+        end
+
+        before do
+          allow_any_instance_of(Graphiti::Util::ValidationResponse)
+            .to receive(:validate!).and_raise(Graphiti::Errors::ValidationError, validation_response)
+        end
+        j
```
richmolj: The letter `j` is randomly here? 😆

spec/integration/rails/commit_hooks_spec.rb
```diff
@@ -0,0 +1,406 @@
...
+        expect {
+          post :create, params: payload
+        }.to raise_error('rollitback')
+        expect(Employee.count).to be_zero
+        expect(Callbacks.entities.length).to eq(1)
+        expect(Callbacks.fired[:create]).to be_a(Employee)
+      end
+
+      context 'when an error is raised before_commit' do
+        it 'does not run after_commit callbacks' do
```
richmolj: 👍

richmolj (approved):
I think this is mergeable, though I had a few comments. Excited to get this in, thanks!

spec/integration/rails/commit_hooks_spec.rb
wadetandy: yeah i was just checking your code review skills...

spec/integration/rails/commit_hooks_spec.rb
wadetandy: updated to prefix all before_commit hooks with `before_`

lib/graphiti/resource_proxy.rb
wadetandy: I've updated `Util::Hooks` to be a more lifecycle aware `Util::TransactionHooksRecorder`.  Instead of returning an implicit array, results are now expected to be returned as a structured `{results:, after_commit_hooks:}` hash.  This isn't quite as clean as I'd like since the `Resource#transaction` method bypasses this structuring when validation errors happen and I need to repeat in the rescue block, and the destroy path doesn't use the hook recorder at all, so some destructuring was needed there, but I think in general this is a lot more readable.

spec/integration/rails/transaction_hooks_spec.rb
```diff
@@ -0,0 +1,406 @@
...
+      {
+        data: {
+          type: 'employees',
+          attributes: { first_name: 'Jane' }
+        }
+      }
+    end
+
+    context 'on create' do
+      it 'fires after validations but before ending the transaction' do
```
richmolj: It's implicit this test is talking about `before_commit` only, can we make explicit/?

lib/graphiti/util/transaction_hooks_recorder.rb
```diff
@@ -0,0 +1,46 @@
+module Graphiti
+  module Util
+    class TransactionHooksRecorder
+      class << self
+        def record
+          reset_hooks
+
+          begin
+            transaction_response = yield.tap do |r|
```
richmolj: I don't have huge objections to this, but if we're going with a pattern that requires a specific payload to be returned, let's add a comment at the top of this class noting that + basic usage.

lib/graphiti/util/transaction_hooks_recorder.rb
```diff
@@ -0,0 +1,46 @@
...
+              run(:before_commit)
+              r[:after_commit_hooks] = hook_set(:after_commit)
```
richmolj, line 37: I personally find this less obvious and more indirection than `Hooks.after_commit_hooks`. But it's not really a problem and an internal API we can change later.

richmolj (approved):
Great to have this in 🎉 !

PR #96, opened 2019-02-21
Prevent query when empty has_many

richmolj:
We were already preventing a sideload query for a belongs_to when the
foreign key was nil. This wasn't done for has_many because we expect you
to always have a *primary* key.

This isn't the case for nonstandard primary keys explicitly specified.
So, let's make sure we prevent a query in this case.

Similarly, both have issues where the pk/fk is nil or unused, but the
parameters are being customized manually. In this case, we want to make
sure we don't accidentally *prevent* the query.

### DIFF
--- a/lib/graphiti/sideload.rb
+++ b/lib/graphiti/sideload.rb
@@ -190,6 +190,7 @@ module Graphiti
     def load(parents, query, graph_parent)
       params = load_params(parents, query)
       params_proc.call(params, parents) if params_proc
+      return [] if blank_query?(params)
       opts = load_options(parents, query)
       opts[:sideload] = self
       opts[:parent] = graph_parent
@@ -325,6 +326,15 @@ module Graphiti
     end
 
     private
+
+    def blank_query?(params)
+      if filter = params[:filter]
+        if filter.values == ['']
+          return true
+        end
+      end
+      false
+    end
 
     def validate_options!(opts)
       if opts[:remote]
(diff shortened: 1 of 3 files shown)


richmolj:
@wadetandy take a look?

lib/graphiti/sideload.rb
```diff
@@ -326,6 +327,15 @@ def resource_class_loaded?
 
     private
 
+    def blank_query?(params)
+      if filter = params[:filter]
+        if filter.values == ['']
```
wadetandy, line 332: This seems both too specific and too general?  What would happen in a scenario where filters = `{foo: '', bar:''}`? 

Should we instead be isolating to something like `params[:filter].has_key?(primary_key) && params[:filter][primary_key].empty?`?

richmolj: > Should we instead be isolating to something like `params[:filter].has_key?(primary_key) && params[:filter][primary_key].empty?`?

We can do this, but we actually catch less cases. That would fix the issue for specifically when graphiti is generating the params via PK/FK. The current implementation would *also* catch

```ruby
has_many :foos do
  params do |hash, model| # model is parent in graph
    hash[:filter] = { id: model.whatever }
  end
end
```

I really don't think we need to worry about all possible edge cases, but this seems reasonably likely to happen, and the current implementation works for this as well as graphiti-generated params.

wadetandy: My main question in the second case would be whether handling this for the user is expected behavior or if understanding "blank" logic should be left to the user in this case.  I lean toward the former as well, but wanted to bring it up.  Fine with merging this if we're in agreement.

richmolj: We are 100% on the same page and that was my initial thought as well. End of the day, I think the above example is just too common to ignore - in fact, I was doing something like this originally when I first encountered the problem.

PR #100, opened 2019-03-03
Fix polymorphic many-to-many

richmolj:
The most common example of this is tags - many different models might
have a tag, and a tag can have many different models. So there's
typically a `taggings` table with `taggable_id` and `taggable_type`.

When sideloading/linking the tags, we were already taking into account
polymorphism. Because M2M is relational-DB-specific, we used
ActiveRecord to introspect the relationships and account for the
polymorphic join.

The problem occurs when you have multiple resources with `many_to_many
:tags` - the filter on `TagResource` would be `taggable_id`
(automatically applied under the hood). But when multiple resources have
this association, they'd continually overwrite the `taggable_id` filter.

The solution is to do this at runtime, not configuration time. When
linking this relationship, `taggable_id` shouldn't be an integer, but a
hash containing both the relevant `id` and the relevant `type`:

`?filter[taggable_id]={"id":1,"type":"authors"}`

For one, we solve the problem and prevent a collision on `taggable_id`.
But we also unlock the behavior of querying by multiple taggables at
once. This type of filter would be able to satisfy something like "give
me all tags for author 123 and book 456".

Notable, the `polymorphic_has_many` relationship queries by passing
*_id and *_type as separate filters. Perhaps this should change, but I
don't think it needs to - the use case is rare, doable with manual
filtering, and the current implementation is a bit more along the lines
of what a user expects.

### DIFF
--- a/lib/graphiti/adapters/active_record/many_to_many_sideload.rb
+++ b/lib/graphiti/adapters/active_record/many_to_many_sideload.rb
@@ -9,20 +9,42 @@ class Graphiti::Adapters::ActiveRecord::ManyToManySideload < Graphiti::Sideload:
   end
 
   def belongs_to_many_filter(scope, value)
-    scope
-      .includes(through_relationship_name)
-      .where(belongs_to_many_clause(value))
+    if polymorphic?
+      clauses = value.group_by { |v| v['type'] }.map do |group|
+        ids = group[1].map { |g| g['id'] }
+        filter_for(scope, ids, group[0])
+      end
+      scope = clauses.shift
+      clauses.each { |c| scope = scope.or(c) }
+      scope
+    else
+      filter_for(scope, value)
+    end
+  end
+
+  def ids_for_parents(parents)
+    if polymorphic?
+      parents.group_by(&:class).map do |group|
+        { id: super(group[1]), type: group[0].name }.to_json
+      end
+    else
+      super
+    end
   end
 
   private
 
-  def belongs_to_many_clause(value)
-    where = { true_foreign_key => value }.tap do |c|
-      if polymorphic?
-        c[foreign_type_column] = foreign_type_value
-      end
-    end
+  def filter_for(scope, value, type = nil)
+    scope
+      .includes(through_relationship_name)
+      .where(belongs_to_many_clause(value, type))
+  end
 
+  def belongs_to_many_clause(value, type)
+    where = { true_foreign_key => value }
+    if polymorphic? && type
+      where[foreign_type_column] = type
+    end
     { through_table_name => where }
   end
 
--- a/spec/integration/rails/finders_spec.rb
+++ b/spec/integration/rails/finders_spec.rb
@@ -4,13 +4,17 @@ if ENV['APPRAISAL_INITIALIZED']
 
     controller(ApplicationController) do
       def index
-        authors = Legacy::AuthorResource.all(params)
-        render jsonapi: authors
+        records = resource.all(params)
+        render jsonapi: records
       end
 
       def show
-        author = Legacy::AuthorResource.find(params)
-        render jsonapi: author
+        record = resource.find(params)
+        render jsonapi: record
+      end
+
+      def resource
+        Legacy::AuthorResource
       end
     end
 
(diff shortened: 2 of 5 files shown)


lib/graphiti/adapters/active_record/many_to_many_sideload.rb
```diff
@@ -9,20 +9,42 @@ def through_relationship_name
...
+    end
   end
 
-  private
-
-  def belongs_to_many_clause(value)
-    where = { true_foreign_key => value }.tap do |c|
-      if polymorphic?
-        c[foreign_type_column] = foreign_type_value
+  def ids_for_parents(parents)
```
richmolj, line 25: This method is used for generating the link/params between resources

lib/graphiti/adapters/active_record/many_to_many_sideload.rb
```diff
@@ -9,20 +9,42 @@ def through_relationship_name
   end
 
   def belongs_to_many_filter(scope, value)
```
richmolj, line 11: this method is for actually scoping based on the incoming params

richmolj:
@wadetandy take a look please?

spec/integration/rails/finders_spec.rb
```diff
@@ -1095,6 +1099,71 @@ def show
...
+            allow(controller).to receive(:resource) { Legacy::BookResource }
+            do_index({ include: 'tags' })
+            sl = d[0].sideload(:tags)
+            expect(sl.map(&:id)).to eq([tag1.id, tag2.id])
+            sl = d[1].sideload(:tags)
+            expect(sl.map(&:id)).to eq([tag1.id])
+          end
+        end
+
+        if ::Rails::VERSION::MAJOR >= 5
```
wadetandy, line 1141: ?

spec/integration/rails/finders_spec.rb
```diff
@@ -1095,6 +1099,71 @@ def show
...
+              book2.tags << tag3
+            end
+
+            it 'still works' do
+              allow(Legacy::TagResource).to receive(:validate_endpoints?) { false }
+              allow(controller).to receive(:resource) { Legacy::TagResource }
+              do_index({
+                filter: {
+                  taggable_id: [
+                    { id: author1.id, type: author1.class.name }.to_json,
```
wadetandy, line 1157: If I'm reading this correctly, the query here would look something like:

```
{ id: 12345, type: 'Legacy::Author' }
```

This is incorrect, IMO.  Instead we should be expecting a jsonapi type and id pair, as the current state exposes too much internal implementation detail.

wadetandy (commented):
The implementation looks good, but see my issue about the new query interface in the specs.

spec/integration/rails/finders_spec.rb
richmolj: Under the hood this uses `.or`, which was introduced in Rails 5

spec/integration/rails/finders_spec.rb
richmolj: It's an interesting point, but not sure if I agree.

Let's say we had a random STI table - no relationships. The ED sample app has `Task` for instance which has a type column with values `Bug`, `Epic`, etc. If you were to query this API, would you expect

`?filter[type]=Bug`

or

`?filter[type]=bugs`

I'd say the former - the `type` here is acting like any other attribute.

So I guess the question here is, since this is a relationship and the JSON is in a particular format, should we treat this specially? I think you can make a case for that, and if anyone wanted to PR with a configuration option I'd support it. But to me the "principle of least surprise", that is most closely following existing usage, is to pass the value explicitly instead of adding this extra indirection.

wadetandy: Yeah I can get behind that from a "what does this do by default?" perspective. I think I probably wouldn't have even thought to say something here if we weren't happening to use a namespaced class in our test code, which reveals more explicitly that we're exposing internal implementation to our API.  

Maybe we merge this, but add a TODO to either the core gem or our cookbooks for a pattern for how to make _both_ of these APIs hide those internal Rails STI/polymorphic associations things away from the outside world.  That way either here or in the simple STI case, the `type` field would be some API-level abstraction and not the AR model we happen to be dealing with.  Not sure where that code _should_ live, in the end.

richmolj: Same page. I think a cookbook would be good for this - even if it's not high-usage, it's a good way to show how to customize things. I've added to my notes, let's get a cookbook section rolling out soon.

PR #101, opened 2019-03-04
Move docs to dedicated repo

wadetandy:
Have removed the documentation files from this project, as they are now hosted in a [dedicated project](https://github.com/graphiti-api/graphiti-api.github.io). This also sets up basic redirects for all of the main landing pages, which should help preserve most bookmarks people might have had.


wadetandy:
@richmolj FYI

PR #102, opened 2019-03-05
Convert codebase to be StandardRB compliant

wadetandy:
As we see an increase in outside contributions, applying a common coding style will be increasingly important.  Rather than spend weeks bikeshedding rubocop rules, I suggest we use the curated ruleset from the TestDouble team, [StandardRB](https://github.com/testdouble/standard).  This PR adds the gem to dependencies and fixes the code to comply over a few steps.

We still need to integrate this with our CI pipeline (holding off due to two failing rules which may also be a bug in the library), and probably figure out a streamlined way to get this to run as a pre-commit hook (I know how to make this work easily in javascript with `husky`, but no idea if ruby has an equivalent).


wadetandy:
@richmolj There are two remaining "UselessAssignment" errors for the unused `opts` variable in

https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/railtie.rb#L59-L62

If I'm reading correctly, we aren't actually doing anything with this.  This is either a bug or a vestigial piece of code that we can get rid of, but wanted to ask you for context.

richmolj:
@wadetandy really fantastic stuff, this has been on the back burner for a while so great to see.

[Here's a guide to pre-commit without a gem](https://github.com/IcaliaLabs/guides/wiki/Ruby-applications-with-git-pre-commit-hook). Personally I'd prefer leaving this up to CI, but I also don't want people pushing PRs then realizing the failure later, so I'd be down.

The `UselessAssignment` errors are actually a good catch of something left over from Suite! We should be doing `options.reverse_merge(default_jsonapi_render_options)` (with a tiny integration spec). This was more common in Suite and quite rare in Graphiti, which is why we haven't seen a problem. Most common use case I can imagine is wanting common `meta` across endpoints (maybe a benchmark or time the request took or something).

lib/generators/graphiti/api_test_generator.rb
```diff
@@ -26,33 +26,33 @@ def var
     end
 
     def dir
-      @resource.gsub('Resource', '').underscore.pluralize
+      @resource.gsub("Resource", "").underscore.pluralize
     end
 
     def generate_api_specs
-      if actions?('index')
+      if actions?("index")
```
richmolj, line 33: surprising to me, I know rubocop wants the opposite

lib/graphiti/delegates/pagination.rb
```diff
@@ -27,11 +27,11 @@ def pagination_link(page)
 
         # Overwrite the pagination query params with the desired page
         uri.query = @proxy.query.hash.merge({
-                                             page: {
```
richmolj, line 30: 💯

richmolj:
Let's make sure to manually validate `tasks.rb` - pretty thin, and light usage, but not caught by specs atm.

lib/generators/graphiti/api_test_generator.rb
wadetandy: From Standard readme:

> Double quotes for string literals - because pre-committing to whether you'll need interpolation in a string slows people down

Makes sense to me.  Always annoying when needing to add interpolation to also need to change out quotes.

wadetandy:
@richmolj I've added standard to the travis config.  I'm going to merge so as not to continually need to rebase as new changes come in.  Have commented out the unused section of the railtie for now to get linting passing, but we should follow up with the fix & specs.

PR #103, opened 2019-03-08
Add appraisal support for Rails 6 beta

wadetandy:
This will probably hit RC around the time of RailsConf, so we should
make sure we're not breaking anything in the leadup to the release.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -1,6 +1,5 @@
 appraise "rails-4" do
   gem "rails", "~> 4.1"
-  gem "jsonapi-rails", "~> 0.3.1", require: "jsonapi/rails"
   gem "rspec-rails"
   gem "sqlite3", "~> 1.3.6"
   gem "database_cleaner"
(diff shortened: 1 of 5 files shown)


Appraisals
```diff
@@ -1,6 +1,5 @@
 appraise "rails-4" do
   gem "rails", "~> 4.1"
-  gem "jsonapi-rails", "~> 0.3.1", require: "jsonapi/rails"
```
wadetandy, line 3: @richmolj Specs pass without this here so I think we can omit?

spec/rails_spec_helper.rb
```diff
@@ -28,6 +28,10 @@ def generate
       config.logger = fake_logger
       Rails.application.routes.default_url_options = {host: "example.com"}
 
+      if Rails::VERSION::MAJOR >= 6
+        Rails.application.config.hosts << "www.example.com"
```
wadetandy, line 32: New security measure in rails 6. From the [changelog](https://github.com/rails/rails/blob/5e6e505083c2b0caf85b2f86c6be3ff3d8750857/railties/CHANGELOG.md):

> The ActionDispatch::HostAuthorization is a new middleware that prevent against DNS rebinding and other Host header attacks. It is included in the development environment by default with the following configuration:

```ruby
Rails.application.config.hosts = [
  IPAddr.new("0.0.0.0/0"), # All IPv4 addresses.
  IPAddr.new("::/0"),      # All IPv6 addresses.
  "localhost"              # The localhost reserved domain.
]
```

Without this setting, request specs fail in rails 6 due to our setting example.com as the default URL host.

PR #104, opened 2019-03-08
Does not work with jsonapi_serializable 0.3.1

wadetandy:
See broken specs


dawidof:
Thank you @wadetandy, I'll have a look on specs

PR #106, opened 2019-03-08
Fix jsonapi-serializable upgrade incompatibility

wadetandy:
Fixes #104


wadetandy:
@richmolj Mind taking a look at this one since you have context on the original overrides?

PR #107, opened 2019-03-09
Accomodate nested json hashes

richmolj:
We could parse simple hashes, and arrays of simple hashes. But a complex
hash, that contains nested hashes within itself, would be incorrectly
parsed - we would think it was an array when it really wasn't. This was
due to a naive implementation splitting on `},{`.

This implementation is more intelligent, only splitting when a full
snippet of JSON is encapsulated. We also avoid attempting to split
altogether when marked `single: true`.


richmolj:
@wadetandy please take a look

wadetandy (approved):
Great and straightforward solution.

Issue #109, opened 2019-03-12
Allow multiple operators on the same attribute while filtering


marcusjanke:
Hey Lee, first of all i need to thank you for this great library! 🥇 thanks for all the effort and time you put into this amazing gem.

My issue is regarding the option to allow multiple operators to be used for filtering on a single attribute. We have the use case, that we have an attribute `:date` and we want to filter all Resources where this date is between a given range of dates. 

I know that i could achieve this, by defining my own custom filter for that, but i'm wondering why it is not supported to do something like 

`MyResource.all({filter: {date: { gte: Date.today, lte: Date.today+2.months }}})` ?

In the code 
https://github.com/graphiti-api/graphiti/blob/fc35e6fe0b9865320390e8c3fe9ceb946fed333c/lib/graphiti/scoping/filter.rb#L82 you are using only the first operator and the corresponding value. Is there any reason why you did it like that and not use all passed operators and values? 

If this would be possible, sure it would allow ambiguous requests to filter for example the same attribute on `eq` and `not_eq` with the same value, which would be kind of stupid.

Would appreciate any help on this topic. If you want, i can also prepare a PR to make this possible.
Best regards 😊 🙏

richmolj:
Hey @marcusjanke, thanks ❤️ !

You can achieve what you want with:

```
?filter[date][gte]=<DATE>&filter[date][lte]=<DATE>
```

What you are thinking with a hash `{ gte: <DATE>, lte: <DATE> }` is possible if you wanted to register a [custom type](https://www.graphiti.dev/guides/concepts/resources#custom-types), but it's not how graphiti works by default.

> If this would be possible, sure it would allow ambiguous requests to filter for example the same attribute on eq and not_eq with the same value, which would be kind of stupid.

You can do this with ActiveRecord, too: `Post.where(id: 1).where(id: 2)`. Ultimately I think it would be nonsensical user error more than anything else.

Does that make sense to you?

richmolj:
I see - thanks for the follow-up explanation. I agree looks like a bug, and also that your solution makes sense. Do you want to try for a PR that fixes things by looping? Would be a big help ❤️

marcusjanke:
@richmolj i've created a PR for this.. would love to have your feedback! 
best regards

PR #110, opened 2019-03-13
Allow an attribute to be filtered with more than one operator

marcusjanke:
fixes the issue: https://github.com/graphiti-api/graphiti/issues/109


spec/integration/rails/finders_spec.rb
```diff
@@ -110,6 +110,14 @@ def resource
                                created_at_date: 1.day.from_now.to_date
       end
 
+      context "when multiple operators" do
```
richmolj, line 113: This is fine to keep here, but ideally would be a PORO spec in `spec/filtering_spec.rb`

spec/scoping/filter_spec.rb
```diff
@@ -0,0 +1,31 @@
+require "spec_helper"
+
+RSpec.describe Graphiti::Scoping::Filter do
```
richmolj: I typically rely more on integration-y specs instead of mocking the order of calls. If we moved this spec to `spec/filtering_spec.rb`, what value would we lose by removing this file?

richmolj (commented):
This is great @marcusjanke, thanks ❤️ 👏 ! Small comments on tests but love this PR!

spec/scoping/filter_spec.rb
mihaimuntenas: @richmolj  I get your point; definitely having integration-y specs makes it easy to maintain when the underlying implementation changes. Though, for specific "core" components within the implementation might help to have unit tests - as this will monitor specific behavior of these components (the "issue" was within the implementation of this class).
Given the actual content of the spec, I would say that, right now, this can be moved as you suggested.
@marcusjanke what do you say?

marcusjanke:
@richmolj thanks for your feedback!
I've removed the `scoping/filter_spec.rb` and added a spec in the `filtering_spec.rb` as you suggested.

elDub:
I have a need for this as well! Great work!

spec/scoping/filter_spec.rb
richmolj: @mihaimuntenas I agree in the broad strokes, especially after we release 1.0. As it stands now, I think the tests in this PR are perfect.

richmolj:
Really great work @marcusjanke this was a big help 👏 ❤️ 🏅 !

richmolj:
Release in `1.0.rc.25`

marcusjanke:
> Really great work @marcusjanke this was a big help 👏 ❤️ 🏅 !

@richmolj you're welcome - glad that i could help!

PR #111, opened 2019-03-15
Render #resolve meta when available

richmolj:
If you wrap the `#resolve` results in a proxy that responds to `meta`,
let's automatically render that for you.


richmolj:
@wadetandy fyi

PR #115, opened 2019-03-24
Wrap common sideload errors

richmolj:
If you already understand how Graphiti works, it's pretty easy to track
down why a sideload error is happening. But if that's not the case, you feel
kinda stuck with a random error. A good example is a nonstandard foreign key,
or forgetting to add an `attr_accessor` to your model.

This commit now wraps sideload errors, so we can provide context around
the place in the lifecycle the error is occurring:

* Parameter building
* Query Building
* Assignment

We'll still contain the original error message and backtrace in the
wrapped error message. This is also compatibile with Error#cause https://www.honeybadger.io/blog/nested-errors-in-ruby-with-exception-cause/

These errors usually only occur during development. Let's give
developers better information.

### DIFF
--- a/lib/graphiti/errors.rb
+++ b/lib/graphiti/errors.rb
@@ -10,7 +10,7 @@ module Graphiti
       end
 
       def message
-        <<~MSG
+        <<-MSG
           The adapter #{@adapter.class} does not implement method '#{@method}', which was requested for attribute '#{@attribute}'. Add this method to your adapter to support this filter operator.
         MSG
       end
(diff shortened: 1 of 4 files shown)


richmolj:
@wadetandy please take a look

lib/graphiti/errors.rb
```diff
@@ -216,6 +216,74 @@ def message
...
+          #{@resource_class.name}: error occurred while sideloading "#{@sideload_name}"!
+
+          The error was raised while attempting to assign relevant model instances. Read
+          more about sideload assignment here: www.graphiti.dev/guides/concepts/resources#customizing-assignment
+
+          A good way to debug is to put a debugger within the 'assign' block.
+
+          Here's the original, underlying error:
+
+                  #{cause.class.name}: #{cause.message}
```
wadetandy: weird indentation

wadetandy (approved):
Looks good. Great change!

PR #118, opened 2019-03-27
Don't link unexpected polymorphic type

richmolj:
If you try to sideload a `polymorphic_belongs_to`, and we encounter a
type not registered (and not in only/except), we'll raise a helpful
error. But if you aren't sideloading, and just generating a Link, and we
encounter an unknown type, you get a really confusing error.

This instead just omits the Link. I don't think we should blow up in
this case because the client isn't actually requesting anything here -
maybe the client only sideloads on data where the types are known.

### DIFF
--- a/lib/graphiti/util/link.rb
+++ b/lib/graphiti/util/link.rb
@@ -11,6 +11,7 @@ module Graphiti
             @sideload = @sideload.children.values.find { |c|
               c.group_name == type.to_sym
             }
+            @polymorphic_sideload_not_found = true unless @sideload
           else
             @linkable = false
           end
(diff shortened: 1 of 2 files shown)


wadetandy (approved):
Specific functionality looks good. I wonder if there's some sort of flag we are missing where we could opt in to blowing up earlier in the process in the case of an unexpected type? It would be helpful for me to be able to opt into the fact that I hadn't defined a mastercard subtype in most cases.  Maybe something like:

```ruby
Graphiti.configure do |c|
  c.raise_error_on_unknown_polymorphic_type = !Rails.env.production?
end
```

richmolj:
I'm OK adding the option in a separate commit, but this seems like the best default

PR #119, opened 2019-03-27
Add "enum" type

wadetandy:
When are dealing with an enum type, we don't want to accept anything not
in the list of values, nor do we want to niavely run them through the
string casting logic that exists in things like the activerecord
case-insensitive string filters.  When using the new `enum` type in
either your filters or your attributes, you will get an error if you
don't provide an `allow` option:

```ruby
# raises Graphiti::Errors::MissingEnumAllowList
attribute :status, :enum

# OK
attribute :status, :enum, allow: ['published', 'draft']
```

Note that this currently only works for filtering.  Eventually, I would
like to add the ability to prevent anything except for the allowlist
from being written automatically as well, but since users should be
handling this at their model or database layer anyway, I'm not worried
about it for this first version.

### DIFF
--- a/lib/graphiti/errors.rb
+++ b/lib/graphiti/errors.rb
@@ -165,6 +165,29 @@ module Graphiti
       end
     end
 
+    class MissingEnumAllowList < Base
+      def initialize(resource_class, filter_name)
+        @resource_class = resource_class
+        @filter_name = filter_name
+      end
+
+      def message
+        <<-MSG
+          #{@resource_class.name} You declared an attribute or filter of type "enum" without providing a list of permitted values, which is required.
+
+          When declaring an attribute:
+
+          attribute :status, :string_enum, allow: ['published', 'draft']
+
+          When declaring a filter:
+
+          filter :status, :string_enum, allow: ['published', 'draft'] do
+            # ...
+          end
+        MSG
+      end
+    end
+
     class InvalidLink < Base
       def initialize(resource_class, sideload, action)
         @resource_class = resource_class
--- a/lib/graphiti/errors.rb
+++ b/lib/graphiti/errors.rb
@@ -166,22 +166,23 @@ module Graphiti
     end
 
     class MissingEnumAllowList < Base
-      def initialize(resource_class, filter_name)
+      def initialize(resource_class, filter_name, enum_type)
         @resource_class = resource_class
         @filter_name = filter_name
+        @enum_type = enum_type
       end
 
       def message
         <<-MSG
-          #{@resource_class.name} You declared an attribute or filter of type "enum" without providing a list of permitted values, which is required.
+          #{@resource_class.name} You declared an attribute or filter of type "#{@enum_type}" without providing a list of permitted values, which is required.
 
           When declaring an attribute:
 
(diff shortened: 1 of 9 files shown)


lib/graphiti/errors.rb
```diff
@@ -165,6 +165,29 @@ def message
...
+        <<-MSG
+          #{resource_class.name} You declared an attribute or filter of type "enum" without providing a list of permitted values, which is required.
+
+          When declaring an attribute:
+
+          attribute :status, :enum, allow: ['published', 'draft']
+
+          When declaring a filter:
+
+          filter :status, :enum, allow: ['published', 'draft'] do
```
richmolj: Nice

lib/graphiti/resource/dsl.rb
```diff
@@ -9,6 +9,12 @@ def filter(name, *args, &blk)
           type_override = args[0]
 
           if (att = get_attr(name, :filterable, raise_error: :only_unsupported))
+            type = type_override || att[:type]
+
+            if type == :enum && opts[:allow].blank?
```
richmolj: Look in the filter code for `canonical` and go that way. The issue here is you might create a type off of enum, like `foo_enum` - you want it to go through all the same enum logic, but with this new type name. That's why types have a `canonical_name`.

lib/graphiti/types.rb
```diff
@@ -101,6 +101,13 @@ def self.map
             kind: "scalar",
             description: "Base Type. Like a normal string, but by default only eq/!eq and case-sensitive.",
           },
+          enum: {
```
richmolj: What about integer enum?

richmolj (approved):
Looks great 👍 !

lib/graphiti/types.rb
wadetandy: are you saying add a second type for that?

richmolj: I think either this is a baseline enum that can work for both strings and integers (considering you're not doing case sensitivity anyway), or it should be a `string_enum` type.

lib/graphiti/resource/dsl.rb
wadetandy: If you had a `string_enum` and `integer_enum` type, would you expect them to both have a `canonical_name == :enum`?

richmolj: Yes, exactly

richmolj: That is why, even if we don't do anything with integers in this PR, we should check the canonical name

wadetandy:
@richmolj Updated to break out `string_enum` and `integer_enum` + `canonical_name`.  Error message also reflects which one was actually used when providing examples.

richmolj:
Code looks perfect, only thing I can think of is maybe explicitly test string vs integer still works (imagine refactoring the code to not use canonical name by accident). Great work!

wadetandy:
Good call. Updated and added a section to the docs: https://www.graphiti.dev/guides/concepts/resources#enum-types

Will merge and cut a new version once specs pass.

richmolj:
Really awesome, thanks for updating the docs 👍

PR #121, opened 2019-03-30
Improve remote resource behavior

richmolj:
* Previously, we'd disallow all writes. Now we allow associating to an
existing remote entity (only applies to `belongs_to).
* Previously, we'd avoid links for remote resources. Now, we avoid link
validation (because we can't validate a remote API), but still generate
the links (you can of course always turn this off).
* As part of the last point, improved foreign key inferrence. We usually
infer FK via the model name, but for remote resources the model is
OpenStruct so we should use the resource name instead.


richmolj:
@wadetandy heads up

PR #123, opened 2019-04-04
Throw better error when polymorphic type not found

richmolj:
If you are sideposting a polymorphic relationship, and pass a jsonapi
type that is not registered, we would throw a confusing error. Now we
throw an explicit error with information about what's happening.

### DIFF
--- a/lib/graphiti/errors.rb
+++ b/lib/graphiti/errors.rb
@@ -561,6 +561,28 @@ module Graphiti
       end
     end
 
+    class PolymorphicSideloadTypeNotFound < Base
+      def initialize(sideload, name)
+        @sideload = sideload
+        @name = name
+      end
+
+      def message
+        <<~MSG
+          #{@sideload.parent_resource}: Tried to find a Resource with type '#{@name.inspect}', but did not find one!
+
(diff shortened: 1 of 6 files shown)


richmolj:
@wadetandy

lib/graphiti/errors.rb
```diff
@@ -561,6 +561,28 @@ def message
...
+      def message
+        <<~MSG
+          #{@sideload.parent_resource}: Tried to find a Resource with type '#{@name.inspect}', but did not find one!
+
+          This is because either a Resource with that type doesn't exist, or it's not registered on the sideload. The below example shows how to register a Resource with this sideload. Make sure one of the registered Resources has type '#{@name.inspect}'
+
+          polymorphic_belongs_to #{@sideload.name.inspect} do
+            group_by(#{@sideload.grouper.field_name.inspect}) do
+              on(:foo)
+              on(:foo).belongs_to :foo, resource: FooResource (long-hand example)
```
wadetandy: Pedantic fix: `on(:foo).belongs_to :foo, resource: FooResource # (long-hand example)`

wadetandy (approved):
Small comment but looks good overall!

Issue #124, opened 2019-04-10
Assumed `payload` presence breaks #destroy


mihaimuntenas:
If the underlying model cannot be destroyed because of validations (eg: ActiveRecord), but rather errors are set, the rendering of :jsonapi_errors fails with:
`ERROR: NoMethodError: undefined method 'relationships' for nil:NilClass`.

This happens because presence of `payload` on the proxy is assumed (which is true for update scenarios), but not for #destroy scenarios.
https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/railtie.rb#L77

A #destroy action requires no additional payload, but rather only the `:id` of the object to be deleted. Thus, the payload is set to `nil`, by the `Base#deserialized_params`
https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/base.rb#L42

@richmolj Any idea which way to fix would be safer/better?
- change `deserialized_params` to return an empty object
- update renderer registration code to permit `nil` payload

richmolj:
Thanks for the well-written issue! ❤️ 

> change deserialized_params to return an empty object

This makes the most sense to me, so we can avoid the conditionals later on.

mihaimuntenas:
@richmolj will be back with a PR.

In the meantime, as a quick fix, one can use the below code in the initialization part of their app,
to replace the railtie code with a patched version:
```ruby
ActiveSupport.on_load(:action_controller) do
  ::ActionController::Renderers.remove(:jsonapi_errors)
  ::ActionController::Renderers.add(:jsonapi_errors) do |proxy, _options|
    self.content_type ||= Mime[:jsonapi]

    validation = GraphitiErrors::Serializers::Validation.new proxy.data, proxy.payload&.relationships

    render json: { errors: validation.errors }, status: :unprocessable_entity
  end
end
```

Note the safe navigation operator on `payload`. Depending on ruby version, this can be replaces with `.try`.

mihaimuntenas:
@richmolj Would you be ok with assuming a default value for the `payload` param of the Deserializer initializer ? (`def initialize(payload = {})`)

https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/deserializer.rb#L48

Asking for a friend 😉 ...

richmolj:
Yes!

PR #125, opened 2019-04-11
Fix Issue #124

mihaimuntenas:
Base#deserialized_params to always return a Deserializer object

### DIFF
--- a/lib/graphiti/base.rb
+++ b/lib/graphiti/base.rb
@@ -41,6 +41,8 @@ module Graphiti
         payload = normalized_params
         if payload[:data] && payload[:data][:type]
           Graphiti::Deserializer.new(payload)
+        else
+          Graphiti::Deserializer.new
         end
       end
     end
--- a/lib/graphiti/deserializer.rb
+++ b/lib/graphiti/deserializer.rb
@@ -45,7 +45,7 @@
 #
 #   { type: 'authors', method: :create, temp_id: 'abc123' }
 class Graphiti::Deserializer
-  def initialize(payload)
+  def initialize(payload = {})
     @payload = payload
     @payload = @payload[:_jsonapi] if @payload.key?(:_jsonapi)
   end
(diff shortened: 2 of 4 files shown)


wadetandy:
@mihaimuntenas This looks like a good fix for the issue.  Would you mind adding a spec (probably makes sense in the `spec/integration/rails` section) which validates the fix? I'm doing some refactoring in this area soon for a separate feature and I don't want to accidentally rebreak this when I do.

mihaimuntenas:
@wadetandy Will check how I can hook into the existent one, from:
https://github.com/graphiti-api/graphiti/blob/master/spec/integration/rails/persistence_spec.rb#L140

Really strange that this one actually did not fail in the first place.

mihaimuntenas:
@wadetandy the issue was caused by the following line of code:
https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/railtie.rb#L77

Any suggestion where I can place a spec for this ?
Will also search on my own, but any suggestion would be appreciated.

mihaimuntenas:
@wadetandy I suppose this is the cause that the actual spec did not catch this:

https://github.com/graphiti-api/graphiti/blob/master/spec/support/rails/employee_controller.rb#L38

this line should read ` render jsonapi_errors: employee` to actually trigger the rendering block setup by the railtie.

Would it be ok to just update this line ?

mihaimuntenas:
PS: same applies generally for tendering the errors in this mock controller (for update and create).

mihaimuntenas:
@wadetandy Updated the specs for persistence in specs/rails/integration. Switched all error rendering to use the `render jsonapi_errors: employee`, relying on graphiti_errors implementation.
Can confirm, that without the fix, the #destroy fails with the issue, as expected.

richmolj (commented):
This looks good to me, great work @mihaimuntenas! 👏 

@wadetandy all good for you?

wadetandy:
Looks great!

richmolj:
Released in `1.0.2`

PR #126, opened 2019-04-16
Request errors

wadetandy:
This is the first phase of work to allow us to provide better & more
actionable errors to API users who might not also be developers of the
API itself.  Previously when trying to write unwritable attributes &
relationships, on typecast problems, or trying to write an attribute
that doesn't exist, the request would blow up with a 500 and an
unhelpful error message.  The request validator takes on most of the
work of transforming and validating an incoming request for processing
all while rescuing any of the above errors and turning them into helpful
error messages.  Note that this is only currently intended to process
write payloads, but will be fully extendable to validate query
parameters for read requests too.

This also adds a new "payload_path" key to the meta payload for the
deserialized/normalized request payload, which allows consumers to know
which path within the request json structure they are dealing with at
any given time. Here it is used for more useful error messages.


wadetandy:
@richmolj Would like to get your feedback on the approach before I drive across the finish line.  The current implementation would replace the typecasting & writability checks currently in `Util::Persistence` and the payload normalization currently in the runner. This would likely be integrated somewhere in the ResourceProxy or Runner layer, but open to other suggestions.

wadetandy:
@richmolj This should be good to go.  Going to follow up with changes to graphiti_errors later this weekend/early next week so that we can render nice errors back to API consumers.

lib/graphiti/runner.rb
```diff
@@ -1,18 +1,27 @@
 module Graphiti
   class Runner
     attr_reader :params
-    include Graphiti::Base
```
richmolj, line 4: Agree moving from `Base` to `Runner` is a good move. Just to be clear, you moved things but did not refactor correct?

lib/graphiti/runner.rb
```diff
@@ -1,18 +1,27 @@
...
 
     def initialize(resource_class, params, query = nil)
       @resource_class = resource_class
       @params = params
       @query = query
+
+      validator = RequestValidator.new(jsonapi_resource, params)
+
+      if validator.validate
```
richmolj: maybe this should be `validate!` which raises the error?

lib/graphiti/runner.rb
```diff
@@ -1,18 +1,27 @@
...
+      else
+        raise Graphiti::Errors::InvalidRequest, validator.errors
+      end
     end
 
     def jsonapi_resource
       @jsonapi_resource ||= @resource_class.new
     end
 
+    attr_reader :deserialized_payload
```
richmolj: top of file please

lib/graphiti/util/persistence.rb
```diff
@@ -180,12 +164,8 @@ def process_has_many(relationships, caller_model)
       iterate(except: [:polymorphic_belongs_to, :belongs_to]) do |x|
         yield x
 
-        if x[:sideload].writable?
```
richmolj, line 183: because this check happens upstream in the validator right?

lib/graphiti/util/simple_errors.rb
```diff
@@ -0,0 +1,89 @@
...
+        end
+      end
+
+      def size
+        values.flatten.size
+      end
+      alias count size
+
+      def values
+        messages.select { |key, value|
```
richmolj: How about `messages.values.reject(&:empty?)`?

richmolj (approved):
Looks really solid. A few tiny questions just to double-check my understanding but great PR. Let's also run through the ED sample app to sanity check before merge.

lib/graphiti/runner.rb
wadetandy: Correct

lib/graphiti/util/persistence.rb
wadetandy: As far as I can tell this can't happen in a way that doesn't go through the runner which does the validator check. Let me know if this isn't true for any reason you're aware of.

lib/graphiti/runner.rb
wadetandy: Added a second `validate!` method to `RequestValidator` which moves this logic there.

wadetandy:
Pushed updates.  Let's get [graphiti_errors#5](https://github.com/graphiti-api/graphiti_errors/pull/5) merged and released before this is merged in.

PR #141, opened 2019-05-13
Switch to graphiti-rails

wagenet:
https://github.com/wagenet/graphiti-rails

To Do:
* [x] Merge https://github.com/wagenet/graphiti-rails/pull/24
* [x] Merge https://github.com/wagenet/graphiti-rails/pull/22
* [x] Release rescue_registry
* [x] Write migration docs
* [x] Update graphiti documentation
* [ ] Update graphiti-rails in-repo docs
* [x] Merge this PR
* [x] Release graphiti 1.2
* [x] Update graphiti-rails to graphiti 1.2
* [x] Release graphiti-rails
* [x] Update Application Template
* [ ] Update Employee Directory Tutorial
* [x] Publish release, link to docs


wagenet:
I'm unable to reproduce the spec failure locally.

richmolj:
@wagenet I retried and it passed. I'm not sure why but ruby 2.6 on Rails 5 gets this failure randomly and I can't reproduce locally either. I haven't put any effort into figuring it out yet, focusing on other things, but it's on the radar.

lib/graphiti.rb
```diff
@@ -12,84 +12,10 @@
...
-
-if defined?(Rails)
-  require "graphiti/railtie"
-  require "graphiti/rails"
-  require "graphiti/responders"
-end
-
 module Graphiti
+  # @api private
+  DEPRECATOR = ActiveSupport::Deprecation.new('2.0', 'Graphiti')
```
richmolj, line 16: 😍

lib/graphiti/configuration.rb
```diff
@@ -66,4 +55,6 @@ def with_option(key, value)
       send(:"#{key}=", original)
     end
   end
+
+  DEPRECATOR.deprecate_methods(Configuration, :respond_to, :respond_to=, "Use graphiti-rails's `config.graphiti.respond_to_formats`")
```
richmolj: I'm practically squealing with delight

lib/graphiti/resource/remote.rb
```diff
@@ -38,6 +38,7 @@ def before_resolve(scope, query)
       # Forward all headers
       def request_headers
         {}.tap do |headers|
+          # TODO: Maybe handle this in graphiti-rails
```
richmolj, line 41: Yeah. These and others are definitely quick fixes that would ideally live in graphiti-rails. That said, they don't bother me too much as long as there aren't too many and they remain small. Up to you.

lib/graphiti/responders.rb
```diff
@@ -1,19 +1,5 @@
...
-    def respond_with(*args, &blk)
-      opts = args.extract_options!
-      opts[:location] = nil
-      args << opts
-      super(*args, &blk)
-    end
-  end
+module Graphiti
+  include ActiveSupport::Deprecation::DeprecatedConstantAccessor
+  deprecate_constant "Responders", "Rails::Responders", message: "Use Graphiti::Rails::Responders from the graphiti-rails gem", deprecator: DEPRECATOR
```
richmolj: ❤️

richmolj (commented):
Looks absolutely fantastic. I'd like to pull down myself and run against some existing APIs, but the code changes here are 💯 👏 🙌 😍 !

lib/graphiti/resource/remote.rb
wagenet: I left them alone for now since they're small and I don't think it's terrible if they're left behind. That said, I think the long term approach is probably to expose a flag on Graphiti that can get set in graphiti-rails.

wagenet:
I _think_ the failing specs will pass if restarted.

richmolj:
I think we're *almost* done here. I didn't see anything to update in the graphiti-rails README so please update if you had something in mind. I also updated the master branch of ED but didn't go step-by-step.

The only thing is the slightly differing error payloads - instead of `__raw_error__` it's `__details__`, the debug information is missing, and a few other minor details. This is important to fix because A) remote resources and vandal parse that payload B) debug information is helpful. I believe we can fix this with just a RescueRegistry release.

PR #144, opened 2019-05-15
Ensure page size 999 for remote APIs

richmolj:
When sideloading, we typically don't add pagination - except when the
sideload is remote, we get a de facto pagination limit because no
page[size] is specified in the URL.

This passes page[size]=999 by default to overcome this issue. In future
PRs, we may want to execute additional requests to get results past 999.

The downside here is existing specs will break, requiring you add
`page[size]=999`. We'll have to overcome this with documentation.

### DIFF
--- a/lib/graphiti/adapters/graphiti_api.rb
+++ b/lib/graphiti/adapters/graphiti_api.rb
@@ -35,6 +35,8 @@ module Graphiti
       def build_url(scope)
         url = resource.remote_url
         params = scope[:params].merge(scope.except(:params))
+        params[:page] ||= {}
+        params[:page][:size] ||= 999
         params = CGI.unescape(params.to_query)
         url = "#{url}?#{params}" unless params.blank?
         url
--- a/spec/remote_resource_spec.rb
+++ b/spec/remote_resource_spec.rb
@@ -55,7 +55,7 @@ RSpec.describe "remote resources" do
   context "when basic query" do
     it "hits remote endpoint" do
       expect(Faraday).to receive(:get)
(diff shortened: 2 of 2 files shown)


richmolj:
@wadetandy not sure we merge yet but please take a look

lib/graphiti/adapters/graphiti_api.rb
```diff
@@ -35,6 +35,8 @@ def handle_remote_error(url, json)
       def build_url(scope)
         url = resource.remote_url
         params = scope[:params].merge(scope.except(:params))
+        params[:page] ||= {}
+        params[:page][:size] ||= 999
```
wadetandy, line 39: I'd like this to be overridable per-remote resource, but this looks good as a general implementation.

richmolj: What would the override be?

wadetandy: Something like max_page_size? Just in case a particular api has changed their default.

richmolj: So what you can currently do is

```
has_many :positions, remote: 'url' do
  params do |hash|
    hash[:page][:size] = 50
  end
end
```

Which would customize on a per-endpoint basis. I can't think of when you'd want something more than this though, or quite where it would go

wadetandy: Ahh yes I forgot this was possible here. I’m fine with this as long as you are confident this can successfully transition to the eventual “find_in_batches” follow on work.

PR #146, opened 2019-05-16
Improve instrumentation

wagenet:
~Wait for #147.~

We should also document these events.

### DIFF
--- a/lib/graphiti/debugger.rb
+++ b/lib/graphiti/debugger.rb
@@ -188,7 +188,7 @@ module Graphiti
     end
 
     ActiveSupport::Notifications.subscribe \
-      "data.graphiti", method(:on_data)
+      "resolve.graphiti", method(:on_data)
     ActiveSupport::Notifications.subscribe \
       "render.graphiti", method(:on_render)
   end
--- a/lib/graphiti/debugger.rb
+++ b/lib/graphiti/debugger.rb
@@ -122,7 +122,7 @@ module Graphiti
       end
 
       def flush
-        Graphiti.broadcast("debug.flush", {}) do |payload|
+        Graphiti.broadcast(:flush_debug, {}) do |payload|
           payload[:chunks] = chunks
           graph_statements.each do |chunk|
             flush_chunk(chunk)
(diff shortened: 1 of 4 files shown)


lib/graphiti/debugger.rb
```diff
@@ -187,8 +187,8 @@ def flush_chunk(chunk, depth = 0)
     end
 
     ActiveSupport::Notifications.subscribe \
-      "graphiti.data", method(:on_data)
+      "data.graphiti", method(:on_data)
```
wadetandy: Should change `data` out for `resolve` here.

wagenet: Fixed.

lib/graphiti/scope.rb
```diff
@@ -77,7 +77,7 @@ def broadcast_data
         sideload: @opts[:sideload],
         parent: @opts[:parent],
       }
-      Graphiti.broadcast("data", opts) do |payload|
+      Graphiti.broadcast(:"scope.resolve", opts) do |payload|
```
wadetandy: Sorry, one more thing.  `scope` is a bit of internal implementation detail in this case, so if we're expecting external consumption of this instrumentation I'd like to separate it a bit.  In `Debugger#on_data` we break out `sideload` vs `primary`, which I think is more clear. How would you feel about changing to that here?

wagenet: Fine with me :)

wagenet:
@wadetandy looking around it seems like AS::N prefers `EVENT.LIBRARY` format with only two segments. At least that's what Rails does, even in cases where you could easily add another segment (e.g. `write_fragment.action_controller` and `read_fragment.action_controller`). If we stick with that, I guess we should do: `debug_flush.graphiti`, `render.graphiti`, `resolve_primary.graphiti`, `resolve_secondary.graphiti`.

wadetandy:
Yeah that makes sense to me and I'm fine with that if it's what Rails, etc tends to do as well.

richmolj (approved):
LGTM, thanks @wagenet ❤️ !

Issue #122, opened 2019-03-30
Add before_validation hook


richmolj:
Original issue: https://github.com/graphiti-api/graphiti/issues/93

This is helpful to manually add validations that are dependent on the graph, e.g. "the employee must have at least two positions".

A-Boudi:
Hi @richmolj,

I'm interested to work on a PR, I'll be working on this in the weekend. But I'm not sure how I should proceed.

I think that it would be possible to run `before_validation` before `before_commit`. This way all the graph should already be persisted before the validations are run.

Issue #93, opened 2019-02-15
rails model validations depending on related entities can break certain create and update scenarios


jnaglick:
Probably best understood via example:

In the graphiti specs, add the following validation to [the Employee class](https://github.com/graphiti-api/graphiti/blob/master/spec/fixtures/employee_directory.rb#L117):

`validates :positions, length: { minimum: 1, too_short: 'An employee must have at least one position!' }`

This validation will break the [persistence > nested create > creates the objects spec](https://github.com/graphiti-api/graphiti/blob/master/spec/integration/rails/persistence_spec.rb#L352).  *Despite the fact this spec attempts to create an employee with positions*, the error response will indicate "An employee must have at least one position!"

----

This appears to be a consequence of how graphiti creates and associates entities as it processes the payload. Looking at [`Graphiti::Util::Persistence#run`](https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/util/persistence.rb#L46), an entity is attempted to be persisted only once and only using its primary attributes. Only after attempting to persist the entity are its associations (parents and children) processed. This means that model validations depending upon parents and/or children do not work as expected. 

I've only given the most simple example here but this bug can manifest in different ways depending upon the exact scenario (i.e. the exact validations, relationships, and payload)

jnaglick:
I have a good deal of background on this and can probably give some more information and/or help fixing, just didn't want to add too much info to the description

richmolj:
Hey John, I know we spoke on this in Slack, is it still outstanding?

leondmello:
I understand you suggested a workaround for the library to take into account validations when saving the record. Can you post it here?

richmolj:
My suggestion was to use the `before_commit` [hook](https://www.graphiti.dev/graphiti/guides/concepts/resources#persistence-lifecycle-hooks), which fires after validations but before committing the transaction. Something like

```ruby
# in model
validates :positions,
  length: { ... },
  on: :before_commit,
  strict: true

# in resource
before_commit do |model|
  model.valid?(:before_commit)
end
```

That's pseudo-code, but should give you an idea.

richmolj:
OK looking deeper into this @jnaglick ...

So the way this works is

* save
* process/associate relationships
* look at `.errors` across all objects and render the validation payload
* `before_commit` hook

We do this because A) we want to decouple from ActiveRecord and B) children need a foreign key to the parent, so we need to save the parent first in order to get that key generated. I think this flow is correct and should not change.

[...]

PR #153, opened 2019-05-26
Add before_validation hooks to resources

A-Boudi:
Solves #122 

This adds a before_validation hook to be fired just before the before_commit hook.

### DIFF
--- a/spec/integration/rails/transaction_hooks_spec.rb
+++ b/spec/integration/rails/transaction_hooks_spec.rb
@@ -1,7 +1,7 @@
 # rubocop: disable Style/GlobalVars
 
 if ENV["APPRAISAL_INITIALIZED"]
-  RSpec.describe "before_ & after_commit hooks", type: :controller do
+  RSpec.describe "before_validation, before_ & after_commit hooks", type: :controller do
     class Callbacks
       class << self
         attr_accessor :fired, :in_transaction_during, :entities
@@ -27,6 +27,7 @@ if ENV["APPRAISAL_INITIALIZED"]
       Callbacks.fired = {}
       Callbacks.in_transaction_during = {}
       Callbacks.entities = []
+      $raise_on_before_validation = {employee: false}
       $raise_on_before_commit = {employee: true}
     end
 
@@ -63,7 +64,7 @@ if ENV["APPRAISAL_INITIALIZED"]
         before_commit do |position|
           Callbacks.add(:before_position, position)
           if $raise_on_before_commit[:position]
-            raise "rollitback_book"
+            raise "rollitback_position"
           end
         end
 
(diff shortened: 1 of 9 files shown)


spec/integration/rails/transaction_hooks_spec.rb
```diff
@@ -230,13 +238,31 @@ def json
...
         end
       end
+
+      context "when an error is raised before_validation" do
+        before do
+          $raise_on_before_validation = {employee: true}
+        end
+
+        it "does not run before_commit callbacks" do
+          expect_any_instance_of(Graphiti::Util::ValidationResponse)
```
richmolj, line 256: You did a great job with the general setup, but I think this is the issue that would hold up the PR. We wanted something to fire ***before*** validation. As you can see and line 126 of `resource_proxy.rb`, this is happening ***after*** validation. IOW, we want `to_not receive` - and probably a Rails integration spec that addresses the [original issue](https://github.com/graphiti-api/graphiti/issues/93).

I don't think you're too far off though! Hope we get this in 👏 !

A-Boudi:
Hi, @richmolj, I'm not sure if I did understand the issue correctly.

Here is what I originally understood:
During the persistance of a resource, graphiti works like this:
* Process resource's parents (mainly belongs_to relationships)
* Persist the resource with the foreing keys retrieved above
* Process resource's children (has_many and many_to_many relationships). Persist the children using the resource key.

If a resource has validation on its children, it can't be saved because the children are not accessible when the resource is being saved.

To fix this issue, the validation on children are scoped via a `:whatever` context. And a second round of validation is done using `:whatever` after that all the resource's parents and children are persisted and accessible.

Now, if we want something to fire before **validation**. Would this mean that this new callback should be fired before `before_save` callback?

A-Boudi:
Hi @richmolj, what do you think?

richmolj:
@A-Boudi thanks for the reminder and sorry for the delay. I think you have correct understanding of the issue, but the word `validation` is tripping things up. Graphiti "validation" is actually just checking model, instances for `#errors` - we expect the actual validation to be done with ActiveModel and performed during save.

So, this is maybe this is better named as `before_validation_processing` or `after_graph_persisted` or something like that.

Still, it should fire before `Graphiti::Util::ValidationResponse` code fires. Otherwise we'd be performing the validation, but nothing would happen in response (no rolling back the transaction or rendering errors correctly).

A-Boudi:
Hi @richmolj,

Like asked, I moved this hook to fire before graphiti validation. I also changed the name to `after_graph_persist`.

For the tests, I have kept the test introduced in the first commit. And I added new integration tests that addresses the [original issue](#93 ).

richmolj:
Really, ***REALLY*** great job @A-Boudi 💯 🥇 ! Thanks so much for sticking with it - I really appreciate the integration tests in particular ❤️  If you could please add an entry to the CHANGELOG, I'd be happy to merge!

A-Boudi:
@richmolj CHANGELOG updated!

PR #157, opened 2019-06-11
Allow attribute to be used, but not to show up in the schema.

zeisler:
Using attribute option `schema: false`.
This option is default true and is not effected by `only` and
`except` options.

### DIFF
--- a/lib/graphiti/schema.rb
+++ b/lib/graphiti/schema.rb
@@ -131,7 +131,7 @@ module Graphiti
     def attributes(resource)
       {}.tap do |attrs|
         resource.attributes.each_pair do |name, config|
-          if config.values_at(:readable, :writable).any?
+          if config.values_at(:readable, :writable).any? && config[:schema]
             attrs[name] = {
               type: config[:type].to_s,
               readable: flag(config[:readable]),
@@ -166,6 +166,8 @@ module Graphiti
     def sorts(resource)
       {}.tap do |s|
         resource.sorts.each_pair do |name, sort|
+          next unless resource.attributes[name][:schema]
+
           config = {}
           config[:only] = sort[:only] if sort[:only]
           attr = resource.attributes[name]
@@ -180,6 +182,8 @@ module Graphiti
     def filters(resource)
       {}.tap do |f|
         resource.filters.each_pair do |name, filter|
+          next unless resource.attributes[name][:schema]
+
           config = {
             type: filter[:type].to_s,
             operators: filter[:operators].keys.map(&:to_s),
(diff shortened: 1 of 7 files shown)


zeisler:
Let me know if you want a PR for docs?

lib/graphiti/schema.rb
```diff
@@ -166,6 +166,8 @@ def flag(value)
     def sorts(resource)
       {}.tap do |s|
         resource.sorts.each_pair do |name, sort|
+          next unless resource.attributes[name][:schema]
```
richmolj, line 169: This seems fine to me, unless you can envision a scenario where we allow something like the sort in the schema but not filtering/reading/etc.

spec/schema_spec.rb
```diff
@@ -185,7 +185,7 @@ def self.name
         self.description = "An employee of the organization"
 
         attribute :first_name, :string, description: "The employee's first name"
-
+        attribute :hidden_attribute, :string, schema: false
```
richmolj, line 188: I know we have the diff spec, but should probably be one here too that flips the boolean and asserts it's now present.

spec/schema_diff_spec.rb
```diff
@@ -240,6 +240,20 @@ def self.name
       it { is_expected.to eq([]) }
     end
 
+    context "when attribute goes schema true to false" do
```
richmolj, line 243: 👍

richmolj (approved):
Great work @zeisler and thanks for this PR ❤️ 

This seems mergeable as-is, but want to note the scenario where we hide based on functionality rather than attribute. Something like:

```ruby
attribute :foo, :string, schema: false
sort :foo, schema: true
```

Not strictly necessary, but something to think about. Up to you.

Something for the docs would be great - ideally there is a separate "Schema" section of the docs where we could put this, but I haven't had time to write that yet. Where would this have been most helpful for you?

lib/graphiti/schema.rb
zeisler: My only use case is that the whole attribute with it's sorting and filtering be hidden. My current project has a table with dynamic columns that changes with customer and year to year. So they are not relevant or available in different contexts. The place where this is hurting is in tests where I create test data that gets added to the schema, which never makes sense in production.

spec/schema_spec.rb
zeisler: Added spec

zeisler:
A schema section does seem helpful in general, but I would go looking for this flag in the resource concepts section. Here what I came up with https://github.com/graphiti-api/graphiti-api.github.io/pull/12

richmolj:
I think this looks great! @zeisler if you could please add to `CHANGELOG` I'll go ahead and merge. Thanks again!

zeisler:
@richmolj changelog added.

richmolj:
Awesome, well done @zeisler

Issue #142, opened 2019-05-15
Allow "null" to be passed in filters


richmolj:
And convert to Ruby's `nil`

richmolj:
Did some initial work here: https://github.com/richmolj/graphiti/tree/filter_nil

Though the work is pretty straightforward, there is a purpose to disallowing nils. Passing a nil can sometimes cause slow/expensive database queries, and it should probably be disabled by default. This means we probably need 2 options

* `filter :foo, allow_nil: true`
* `self.filters_accept_nil_by_default = true (`defaults to false`)

@wadetandy sound about right?

wadetandy:
Makes sense and I agree that we should not change the defaults.  If we're doing some nice things on behalf of `nil`, it might be a good idea to do the same for javascript's `undefined`, though I can also see this being a more common thing that is actual a desired filter.  Maybe have a configurable list of "strings we coerce to `nil`" with sensible defaults?

zeisler:
I would recommend following the JSON spec that `undefined` is an invalid conversion to Ruby.

```ruby
> JSON.parse("undefined")
  #=> JSON::ParserError (416: unexpected token at 'undefined')
> JSON.parse("null")
  #=> nil
```

richmolj:
Yeah, I agree with @zeisler - also makes sense because in Spraypaint, I would drop `undefined` and not send as part of the request, but `null` should turn into the string `"null"`.

Maybe we'll move to a configurable whitelist at some point, but that seems good for now.

wadetandy:
In general I would Be in favor of either allowing nil and undefined OR allowing neither, but I’m not going to get too worked up about any outcome here.

jkeen:
I think this can be closed. Filters accept null → nil via allow_nil on a filter, or globally with filters_accept_nil_by_default

PR #158, opened 2019-06-11
Option to coerce "null" to Ruby nil

zeisler:
Filters have new options `allow_nil: true`
Option can be set at the resource level `Resource.filters_accept_nil_by_default = true`
By default this is set to false.

#142

### DIFF
--- a/lib/graphiti/scoping/filter.rb
+++ b/lib/graphiti/scoping/filter.rb
@@ -54,6 +54,7 @@ module Graphiti
           unless type[:canonical_name] == :hash || !value.is_a?(String)
             value = parse_string_value(filter.values[0], value)
           end
+          value = parse_string_null(filter.values[0], value)
           validate_singular(resource, filter, value)
           value = coerce_types(filter.values[0], param_name.to_sym, value)
           validate_allowlist(resource, filter, value)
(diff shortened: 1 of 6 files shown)


lib/graphiti/scoping/filter.rb
```diff
@@ -53,6 +53,7 @@ def each_filter
           type = Types[filter.values[0][:type]]
           unless type[:canonical_name] == :hash || !value.is_a?(String)
```
richmolj, line 54: Seems like this would not apply to hash types - would it be possible to move out of this block so all types could support `null`?

richmolj (commented):
I love you for this PR ❤️ 😃 ! Thank you!

My only thought is validating this will work for other types than strings (particularly hash, though not the biggest deal). Seems like it would, but maybe a test on `age` would be helpful?

zeisler:
@richmolj changes made.

richmolj:
Great stuff 🎉 ! If you could add to `CHANGELOG` I'm ready to merge + associated docs. Thank you!

zeisler:
@richmolj change log added.

richmolj:
Thanks @zeisler, you rock ❤️ There's a conflict I can't fix though GH atm, but if you can resolve I'm ready to merge.

zeisler:
@richmolj resolved.

PR #159, opened 2019-06-12
Add hook for attribute lookup `attribute_missing` method.

zeisler:
Defining the instance method `attribute_missing(attr)`
allows attributes to be added at runtime.

```ruby
  def attribute_missing(attr)
    self.class.attribute attr, :string if attr == :dynamic_attribute
  end
end
```

### DIFF
--- a/lib/graphiti/resource/interface.rb
+++ b/lib/graphiti/resource/interface.rb
@@ -3,6 +3,9 @@ module Graphiti
     module Interface
       extend ActiveSupport::Concern
 
+      def attribute_missing(attr)
+      end
+
       class_methods do
         def all(params = {}, base_scope = nil)
           validate!(params)
--- a/lib/graphiti/util/attribute_check.rb
+++ b/lib/graphiti/util/attribute_check.rb
@@ -17,37 +17,18 @@ module Graphiti
       end
 
       def run
-        if attribute?
-          if supported?
-            if guarded?
-              if guard_passes?
-                attribute
-              else
-                maybe_raise(request: true, guard: attribute[flag])
-              end
-            else
-              attribute
-            end
-          else
-            maybe_raise(exists: true)
-          end
-        else
-          maybe_raise(exists: false)
-        end
+        return attribute if guard_check! && attribute_check! && supported_check!
+
+        false
       end
 
       def maybe_raise(opts = {})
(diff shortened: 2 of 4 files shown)


zeisler:
@richmolj the failure on travis does not look related to my change.

lib/graphiti/resource/interface.rb
```diff
@@ -50,6 +50,11 @@ def validate!(params)
           end
         end
       end
+
+      included do
```
richmolj: Is there a reason for the `included` hook, as opposed to just `def`? Should add as an instance method either way?

lib/graphiti/util/attribute_check.rb
```diff
@@ -65,7 +65,10 @@ def supported?
       end
 
       def attribute
-        @attribute ||= resource.all_attributes[name]
+        @attribute ||= resource.all_attributes.fetch(name) do |attr|
+          resource.attribute_missing(attr)
```
richmolj: I have no real problem with this code, but what do you think about moving to [this line](https://github.com/graphiti-api/graphiti/blob/e3d4136e1c2b96a9f60d36d8bc41e7c0b9e09acb/lib/graphiti/util/attribute_check.rb#L32). The thought being the default logic of `attribute_missing` should hold the code that raises the error. I can see how that might be a PITA with this as a separate class, just want to get your thoughts.

richmolj (commented):
I think this is mergeable, though wanted to run a thought by you. Also if you could add something to the CHANGELOG please.

This functionality is great to see - thanks so much for your help ❤️ !

lib/graphiti/util/attribute_check.rb
zeisler: I reworked this class a bit. It now has series of checks one that's deals with attributes, `attribute_check!` that also calls `Resource#attribute_missing`. I didn't want to put any logic in that template method because that would require user to reimplement it or call to `super` all of which break the open closed principle.

lib/graphiti/resource/interface.rb
zeisler: 👍

zeisler:
Ok, closing in favor of monkey patching `#get_attr` as per our conversations outside of this PR. Thanks for your time @richmolj

PR #161, opened 2019-07-08
Properly escape singular filters

richmolj:
0e2c4ddd05d5188d5 caused singular filters to avoid parsing the comma
into an array. However, it accidentally also prevented escaping special
characters. As a side effect, a `single: true` filter wrapped in `{{}}`
curlies would incorrectly see the value passed in containing `{{}}`.

This fixes things so singular filters still get parsed correctly, they
just avoid the comma/array scenario as originally intended.


richmolj:
@wadetandy

wadetandy (approved):

PR #166, opened 2019-08-08
Remove unnecessary (and incorrect) conditional

wagenet:
Introduced in ef7fa514279b4da6807e5cf554e06988a58bdb52

### DIFF
--- a/lib/graphiti/util/serializer_relationships.rb
+++ b/lib/graphiti/util/serializer_relationships.rb
@@ -58,9 +58,7 @@ module Graphiti
               self_ref.send(:validate_link!) unless self_ref.send(:eagerly_validate_links?)
 
               link(:related) do
-                if self_ref
-                  ::Graphiti::Util::Link.new(sideload_ref, @object).generate
-                end
+                ::Graphiti::Util::Link.new(sideload_ref, @object).generate
               end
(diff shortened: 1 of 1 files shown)


lib/graphiti/util/serializer_relationships.rb
```diff
@@ -58,9 +58,7 @@ def block
               self_ref.send(:validate_link!) unless self_ref.send(:eagerly_validate_links?)
 
               link(:related) do
-                if self_ref
```
zpencerq, line 61: It looks like this should have been:

```ruby
if links_ref
  ::Graphiti::Util::Link.new(sideload_ref, @object).generate
end
```

with L58 becoming:
```ruby
  if (links_ref = @proxy.query.links?)
```

Originally, it was `_links` coming from `_links = @proxy.query.links?` but the reference was broken in ef7fa514279b4da6807e5cf554e06988a58bdb52 as you stated and the assignment was removed due to a lint in 1d4d08316731ecdbd505e25138d8b301c376c8ce.

wagenet: Yup, that's what I saw too. And since we have the conditional for links_ref above we don't need it again here.

PR #173, opened 2019-09-19
Replace `relationship_links_by_default` with `autolink`

bstchow:
The `InvalidLink` error references a `relationship_links_by_default` setter in Graphiti::Resource, which exists nowhere in the library. It seems the intention was to reference the `autolink` property of Graphiti::Resource.


richmolj:
Thanks @bstchow and sorry for the delay!

PR #177, opened 2019-10-07
[bugfix] Allow hashes to have custom filters

### LINKED ISSUE #164: Cannot make custom operators for `hash` type
You cannot make custom operators for a filter of `hash` type. They all get coerced to be `eq`. Looks like its happening on this line: https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/scoping/filter.rb#L83
https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/scoping/filter.rb#L131-L133

It may be this is code is trying to support the syntax of `filter[metadata]={ "foo": 100 }` instead of the regular `filter[metadata][eq]={ "foo": 100 }`

If that's the case, perhaps the `=` character should be targeted and supported specifically instead of coercing all other potential operator types to be `eq`

caseyprovost:
Resolves: #164 

### Description
Sometimes you want to perform other operations on a hash other an equals. Now you can :)

### Background
Previously, Graphiti always defaulted to the `eq` operator for hashes. This appears to have been for convenience sake so that when you do `filter[data]={ foo: bar}` it evaluated against an equals operation, rather than having to type `filter[data][eq]={ foo: bar }`. 

### Approach
The idea here is to only default the operator to `eq` if the current resource and filter combo do not have that operator defined. In an ideal world, we would probably want to never default...but that would break some backward compatibility. Maybe in a couple of minor versions forward? 🙏

### DIFF
--- a/lib/graphiti/scoping/filter.rb
+++ b/lib/graphiti/scoping/filter.rb
@@ -128,7 +128,7 @@ module Graphiti
     end
 
     def parse_hash_value(filter, param_value, value, operator)
-      has_filter = filter.values[0][:operators][operator].present?
+      has_filter = resource.filters.dig(filter.keys.first, :operators, operator).present?
 
       if operator != :eq && !has_filter
         operator = :eq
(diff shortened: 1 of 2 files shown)


richmolj:
Perfect, great work @caseyprovost! Love the test and PR description. Thanks so much ❤️ 🎉 !

richmolj:
Released in `1.2.5`

PR #178, opened 2019-10-08
Features/relationship protection blocks

### LINKED ISSUE #143: Relationship readable flag with guard
For attributes, you can

```ruby
attribute :foo, :string, readable: :admin?

def admin?
  false
end
```

But this does not work the same for relationships - only `true` and `false` are supported.

caseyprovost:
Resolves: https://github.com/graphiti-api/graphiti/issues/143

### Background
Previously, when defining relationships on a resource it was only possible to define `readable` and `writable` as booleans. However, as permissions come into play this quickly becomes insufficient and requires a manual defining of the relationship via the block system. This could be a little nicer :)

### Approach
* Allow symbols to be passed to `readable` and `writable`. These will get evaluated as methods on the resource.
* Allow symbols to be passed to `readable` and `writable`. These will get evaluated as methods on the resource.
* Allow blocks to be passed to `readable` and `writable`. These will get evaluated without resource context.

### DIFF
--- a/lib/graphiti/sideload.rb
+++ b/lib/graphiti/sideload.rb
@@ -26,8 +26,8 @@ module Graphiti
       @foreign_key                   = opts[:foreign_key]
       @type                          = opts[:type]
       @base_scope                    = opts[:base_scope]
-      @readable                      = opts[:readable]
-      @writable                      = opts[:writable]
+      @readable                      = evaluate_flag(opts[:readable])
+      @writable                      = evaluate_flag(opts[:writable])
       @as                            = opts[:as]
       @link                          = opts[:link]
       @single                        = opts[:single]
@@ -410,6 +410,21 @@ module Graphiti
 
     def namespace_for(klass)
       Util::Class.namespace_for(klass)
+    end
+
+    def evaluate_flag(flag)
+      return false if flag.blank?
+
+      case flag.class.name
+      when "Symbol"
+        resource.send(flag)
+      when "Proc"
+        flag.call
+      when "String"
+        resource.send(flag)
+      else
+        !!flag
+      end
     end
   end
 end
--- a/lib/graphiti/sideload.rb
+++ b/lib/graphiti/sideload.rb
@@ -419,7 +419,7 @@ module Graphiti
       when "Symbol","String"
         resource.send(flag)
       when "Proc"
-        flag.call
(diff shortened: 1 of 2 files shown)


lib/graphiti/sideload.rb
```diff
@@ -411,5 +411,20 @@ def infer_resource_class
     def namespace_for(klass)
       Util::Class.namespace_for(klass)
     end
+
+    def evaluate_flag(flag)
+      return false if flag.blank?
+
+      case flag.class.name
```
richmolj, line 418: Pretty minor, but could we avoid the `.name` string allocation? Also `when Symbol, Proc` since they do the same thing.

lib/graphiti/sideload.rb
```diff
@@ -411,5 +411,20 @@ def infer_resource_class
...
+      when "Symbol"
+        resource.send(flag)
+      when "Proc"
+        flag.call
```
richmolj: This isn't evaluating in the context of the resource, you'll need `instance_eval` (and spec that tests it)

richmolj (commented):
Some comments but this is WONDERFUL to see 😍 !

lib/graphiti/sideload.rb
caseyprovost: I'm not sure how to do that successfully, to be honest. When I do that and change the `when` statements to classes like `String`. `Proc`, and `Symbol` the code fails.

lib/graphiti/sideload.rb
caseyprovost: @richmolj I took the approach of passing the resource as the explicit context to the blocks through `instance_exec`. Would love your thoughts/feedback :)

lib/graphiti/sideload.rb
richmolj: weird

lib/graphiti/sideload.rb
richmolj: Works for me!

richmolj (approved):

richmolj:
Released in `1.2.6`

richmolj:
@caseyprovost would you mind updating the docs?

caseyprovost:
@richmolj of course! :) Shall do that right after dinner ❤️🍺

PR #181, opened 2019-10-18
Validate request structure when writing a resource

sideshowbandana:
This change will handle raise a `Graphiti::Errors::ConflictRequest` when a request payload is malformed. According to https://jsonapi.org/format/#crud-updating, a request is invalid if the payload doesn't match the requested endpoint on a `PUT` request.


richmolj (approved):

richmolj:
Thanks so much for this great PR @sideshowbandana ! 🎉 ❤️

PR #185, opened 2019-10-21
Fix unnecessary queries

richmolj:
PR https://github.com/graphiti-api/graphiti/pull/168 introduced N+1 by
default. This removes that as the default option, and we'll look into it
more in a future PR.


richmolj:
@jkeen

jkeen:
@richmolj Where did this introduce N+1? Do you have an example I could step through, or a test case?

richmolj:
@jkeen sorry for the delay, I was in the middle of a dozen things and wanted to push a fix ASAP. Try the sample app with graphiti `1.2.4` and hit `/api/v1/positions` - you'll see that because we're rendering the resource identifier for the department, it's causing an N+1

Issue #186, opened 2019-10-24
Sideloaded Resource objects are re-used across requests


zpencerq:
This is a pretty obscure behavior in Graphiti but appears to be a result of the `Sideload` classes memoizing their internal resource instances. These sideload resources are stored on the `Resource` class itself. You can see them in `MyResource.sideloads`.

As a result, if you have any sort of stateful work (memoization) inside your Resource, then it can cause subtle bugs. (Perhaps you wanted to memoize some display logic computation that's shared across fields?)

-------

#### Background
I'm using Graphiti with Pundit roughly as follows:
```ruby
class ApplicationResource < Graphiti::Resource
  include Pundit

  def base_scope
    policy_scope(self.model)
  end

  def current_user
    context.current_user
  end

  # ...
end
```

#### Behavior
I found that users were being resolved with policy_scopes that were based on previous users.

#### Digging
It turns out that Pundit has internal caches for policies and policy_scopes. 
Essentially, whenever `policy_scope(scope)` is called, it will use a standard `||=` memoization strategy to save computation within a request using the `scope` as the key. This cache is an instance variable. (https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L323). 

This odd re-use of `Pundit` objects doesn't occur with standard Controllers or sideload-less Graphiti because these objects are instantiated anew per-request.

#### My work-around

I was able to circumvent this issue in the short term (by completely bypassing Pundit's cache) by using:
```ruby
  def base_scope
    Pundit.policy_scope!(current_user, self.model)
  end
```

An alternative would be to just do `policy_scope(self.model.all)` but that will unnecessarily bloat the Pundit cache of the sideload class (since the cache key would be new every time). 

### Conclusion

Perhaps this bug is outside the scope of Graphiti to deal with, considering it's due in-part to integrating with another library. However, the slight difference in instantiation between non-sideloaded Resources and sideloaded Resources may cause subtle bugs in developer's applications. 

For what it's worth, my expectation of the Resources was that they would be similar to Rails Controllers in that they would be instantiated per-request.

richmolj:
Hey @zpencerq, thanks for the excellent write-up. TBH I hadn't really come across this before - since resources operate on separate model instances, it seems odd you'd want to memoize...but this is a valid edge case, and there are probably others.

I haven't benchmarked but my hunch is we need to keep the resource memoization within a request to avoid extra object allocations. But we need to clear that memoization when the request completes.

[...]

PR #187, opened 2019-10-25
Clear Sideload resources after context block returns

zpencerq:
This is a fairly straight-forward fix for https://github.com/graphiti-api/graphiti/issues/186 outlined in https://github.com/graphiti-api/graphiti/issues/186#issuecomment-546337841. I'm not sure if there are other transient objects that should be cleared at the same time but this passes the test.

This solution probably isn't the best since you can nest Graphiti contexts and the closing of an inner context will affect the outer context by this change.

### DIFF
--- a/lib/graphiti/sideload.rb
+++ b/lib/graphiti/sideload.rb
@@ -224,6 +224,7 @@ module Graphiti
       @parent_resource ||= parent_resource_class.new
     end
 
+    # See https://github.com/graphiti-api/graphiti/issues/186
     def clear_resources
       @resource = nil
       @parent_resource = nil
--- a/lib/graphiti/sideload.rb
+++ b/lib/graphiti/sideload.rb
@@ -224,6 +224,11 @@ module Graphiti
       @parent_resource ||= parent_resource_class.new
     end
 
+    def clear_resources
+      @resource = nil
+      @parent_resource = nil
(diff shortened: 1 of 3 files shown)


lib/graphiti/sideload.rb
```diff
@@ -224,6 +224,11 @@ def parent_resource
       @parent_resource ||= parent_resource_class.new
     end
 
+    def clear_resources
```
richmolj, line 228: Could we add a link to the original issue for this, just to jog memory when we come across this code later on?

richmolj (approved):
@zpencerq this is great ❤️ ! I appreciate the incremental fix to the immediate issue. I'm curious to see if it pops up again before a more significant refactor. Nested contexts is a great thing to bring up, but something I think I'm OK with since we're only clearing the memoization, not changing anything.

If you could add the comment, I'm ready to merge. Thanks so much 💯 !

lib/graphiti/sideload.rb
zpencerq: Added in 46453e71f33592b50936c88e3db65645d357f1bd

richmolj:
Awesome, thanks @zpencerq ❤️ !

richmolj:
Released in `1.2.9`!

PR #189, opened 2019-10-30
Convert resource type to symbol

juggy:
Coerce resource type to symbol under the hood to avoid types that do not match problems down the road.


richmolj (approved):

richmolj:
Thanks so much @juggy !

PR #190, opened 2019-11-06
Allow updates to process reserved keywords

richmolj:
Params like 'page' would throw an error, even though they were in the
attributes payload. This is because we passed all params to the "find"
call preceding the update. I can't think of a reason we should pass all
parameters here, we only need the `id` - so let's pass only that and fix
the issue.


richmolj:
Thanks @wadetandy !

PR #191, opened 2019-11-06
Support Ruby Hash and deeply nested JSON Objects

zeisler:
Support Ruby Hash type along with JSON String for type hash
* Fixes issue where each key/value pair would call the filter. 
Allow deeply nested JSON Objects when using type `:hash` and `single: true`

Current feature set does not allow for deeply nested JSON Objects when not using `single: true` because values can be escaped ie. `{{{escaped json string}}}`

### DIFF
--- a/spec/filtering_spec.rb
+++ b/spec/filtering_spec.rb
@@ -101,6 +101,29 @@ RSpec.describe "filtering" do
         expect(Graphiti::Util::Hash).to_not receive(:split_json)
         expect(records.map(&:id)).to eq([employee2.id, employee3.id])
       end
+
+      context "and it is a deeply nested hash" do
+        before do
+          resource.filter :by_json, :hash, single: true do
+            eq do |scope, value|
+              ids = [value["users"]["update"]["id"]["a"], value["admins"]["update"]["id"]["b"]]
+              scope[:conditions][:id] = ids
+              scope
+            end
+          end
+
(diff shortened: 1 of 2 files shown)


spec/filtering_spec.rb
```diff
@@ -101,6 +101,29 @@ def self.name
         expect(Graphiti::Util::Hash).to_not receive(:split_json)
         expect(records.map(&:id)).to eq([employee2.id, employee3.id])
       end
+
+      context "and it is a complex hash" do
```
richmolj: Looks like we have two tests with the same description - maybe "deeply nested hash"?

richmolj (approved):
Thanks @zeisler - I know the code processing hash filters is a big wonky, so appreciate you dealing with it and sending this PR ❤️ 

I take your point about deeply nested hashes and escaping. I can think of a few workarounds, but right now I'm tempted to cross that bridge when we come to it (unless you're interested in taking this further). As it stands I think this can be merged after a minor change to the spec description.

Again, thanks so much!

zeisler:
I don't have any use case for using a single:false hash type at the moment. (If I did it does have surprising results.) So I'm happy to go forward with this until that day comes.

spec/filtering_spec.rb
zeisler: Done!

PR #193, opened 2019-11-07
Provide resource hook to mutate scope after filters are applied.

zeisler:
In order to do scope manipulation only after all provided filters are applied adding a hook `#after_filtering(scope)`. The resulting scope will be used for sorting, pagination, and stats. This might be useful for doing a group by or some other one time query where the filtered scope is used as the sub query.

### DIFF
--- a/lib/graphiti/resource.rb
+++ b/lib/graphiti/resource.rb
@@ -20,7 +20,7 @@ module Graphiti
       yield scope
     end
 
-    def after_filters(scope)
+    def after_filtering(scope)
       scope
     end
 
--- a/lib/graphiti/resource.rb
+++ b/lib/graphiti/resource.rb
@@ -20,6 +20,10 @@ module Graphiti
       yield scope
     end
 
+    def after_filters(scope)
+      scope
+    end
+
     def serializer_for(model)
       serializer
     end
--- a/lib/graphiti/scoping/filter.rb
+++ b/lib/graphiti/scoping/filter.rb
@@ -16,7 +16,7 @@ module Graphiti
         @scope = filter_scope(filter, operator, value)
       end
 
-      @scope
+      resource.after_filters(@scope)
     end
 
     private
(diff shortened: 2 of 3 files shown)


richmolj:
I'm not totally against it, but how feasible is it to use the existing `#around_scoping` for this?

zeisler:
The order of operation in `#around_scoping` is `DefaultFilter` -> `Filter` -> `Sort` -> `Paginate`. Where the scope is mutated and pass along at each phase. Then later the scope is passed to `stats`. I need something to effect everything downstream of `Filter` I don't see a way to do that with `#around_scoping`.

richmolj:
Apologies for the delay @zeisler - I think this looks great. My only thing is the `after_filters` name...what about `after_filtering`? It just currently strikes me as too close to "add an 'after filter'" versus "this is a hook that runs after filtering".

lib/graphiti/resource.rb
```diff
@@ -20,6 +20,10 @@ def around_scoping(scope, query_hash)
       yield scope
     end
 
+    def after_filters(scope)
```
zeisler:
```suggestion
    def after_filtering(scope)
```

lib/graphiti/scoping/filter.rb
```diff
@@ -16,7 +16,7 @@ def apply
         @scope = filter_scope(filter, operator, value)
       end
 
-      @scope
+      resource.after_filters(@scope)
```
zeisler:
```suggestion
      resource.after_filtering(@scope)
```

spec/filtering_spec.rb
```diff
@@ -1677,4 +1677,23 @@ def admin?
       end
     end
   end
+
+  context "when calling #after_filters hook" do
+    before do
+      resource.class_eval do
+        def after_filters(scope)
```
zeisler:
```suggestion
        def after_filtering(scope)
```

spec/filtering_spec.rb
```diff
@@ -1677,4 +1677,23 @@ def admin?
       end
     end
   end
+
+  context "when calling #after_filters hook" do
```
zeisler:
```suggestion
  context "when calling #after_filtering hook" do
```

zeisler:
@richmolj I've made the changes you've recommended.

PR #196, opened 2020-01-07
Excape `match` operations to prevent DoS queries

wadetandy:
Previously when doing a `match` operation against an activerecord string
field, we passed down the string directly and surrounded it with `%` on
either side, allowing a substring match of the column.  There was
nothing preventing a client from including additional `%` wildcard
characters in their query string to add more advanced substring
matching. The problem with this is that if an attacker provides a few
dozen wildcard characters, the sql engine can very quickly run into
processing problems and cause a Denial of Service against the
database/application.

This commit takes advantage of activerecord and arel's sql escaping
behavior to make this safe, at the tradeoff of not being able to pass
additional wildcard characters as part of queries.  Note that rails 4.x
has fewer APIs for this behavior and they aren't all public. Because
that is now an unsupported version of rails, this PR simply keeps the
less safe older behavior when that version of activerecord is detected.

### DIFF
--- a/lib/graphiti/adapters/active_record.rb
+++ b/lib/graphiti/adapters/active_record.rb
@@ -79,11 +79,30 @@ module Graphiti
         filter_string_suffix(scope, attribute, value, is_not: true)
       end
 
-      def filter_string_match(scope, attribute, value, is_not: false)
-        column = column_for(scope, attribute)
-        map = value.map { |v| "%#{v.downcase}%" }
-        clause = column.lower.matches_any(map)
-        is_not ? scope.where.not(clause) : scope.where(clause)
+      # Arel has different match escaping behavior before rails 5.
+      # Since rails 4.x does not expose methods to escape LIKE statements
+      # anyway, we just don't support proper LIKE escaping in those versions.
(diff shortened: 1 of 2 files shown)


wadetandy:
@richmolj I realized this was an issue today so put together a preliminary patch.  Don't love having the two different method definitions, but also wanted to discuss whether we even want to still support rails 4.2 anymore before addressing the issue differently, as we can maybe just throw away the older code path.  

Also you have a better sense of how users might be using the `match` operator than I do, I think.  This may break some legitimate use cases where they expose the wildcard operator as a feature of the app search, but I think it's worth it for most cases where an app is opening itself up to a DoS without realizing.

wadetandy:
If we are worried about users who are relying on the older insecure behavior, we could potentially offer a config option:

``` ruby
Graphiti.configure do |c|
  c.allow_match_operation_wildcards = true # default false
end
```

richmolj:
@wadetandy wonderful PR and description. I think this is correct as-is - let's keep the Rails 4 codepath for now and delete it later. I don't know of anyone who currently needs the configuration option, so let's add it when it comes up.

richmolj:
Released in `1.2.13`

PR #197, opened 2020-01-12
Bypass required filters on .find

richmolj:
If there's a required filter, we should only require it when calling
`.all` and not `.find` - because there's no filtering to do here, we
already have the ID. Any auth would already be done via `base_scope`.

The other thing special about `.find` is we should raise an error when a
record is missing so we can render a 404. This follows that same
pattern, passing a special flag for this use case.


richmolj:
@wadetandy @wadetandy could you please take a look?

wadetandy:
Yep makes perfect sense. Good description.

PR #202, opened 2020-01-24
Fix nested resource lookup by type in update requests

piotrekbator:
I noticed that `UpdateValidator` compares type from payload with type from resource by typecasting the latter to string. However for nested resource is being looked up by `Polymorphism.resource_for_type` which expects type to be a symbol. This doesn't work when performed with a user-supplied string type from `UpdateValidator`.

In this PR you'll find both a fix and a test that fails without the fix.


richmolj:
Fantastic, thanks so much @piotrekbator !

Issue #207, opened 2020-02-06
`match` filter doesn't work with rails 5.1(.7)


jackcasey:
The ActiveRecord Adaptor calls to `sanitize_sql_like` which is apparently not a supported API and has been removed at least in Rails 5.1 afaics:

I'm not sure what other versions of rails are affected, we are on 5.1.7

https://github.com/graphiti-api/graphiti/blob/a5c32ed3bb08459b7f52af9d4f2cb5ae74df438e/lib/graphiti/adapters/active_record.rb#L91

Gives:
```
NoMethodError: undefined method `sanitize_sql_like' for #<G::Grower::ActiveRecord_Relation:0x00007fd2b7f80ae8>
```

More info:
https://github.com/rails/rails/issues/28947

It appears like the fix is simple, just need to call out to a class that includes the `ActiveRecord::Sanitization` concern. 

I will try to tackle this at some point but currently just getting started with Graphiti and would need to figure out how to go about setting up a failing test for this to properly make a good PR for the issue. (BTW I'm in love with Graphiti so far, thanks so much for the awesome library).

wadetandy:
@jackcasey I believe I fixed this issue in my pr #208. Would you be able to verify that patch fixes your issue before we release a new version?

jackcasey:
Just for completeness in case you missed it; I replied to that PR #208. It works fine :) Thanks.

jackcasey:
This should be fixed by #219 just forgot to close.

PR #208, opened 2020-02-06
Properly excape prefix and suffix queries

wadetandy:
In #196 I added sanitization for `match` filters but I negelected
to handle `prefix` and `suffix` filters, which share similar logic.
This handles all three.  Additionally, issue #207 indicates an issue
with the `sanitize_sql_like` method not always being available on the
scope object. Wasn't able to reproduce myself, but this switches to
avoid using the method from scope directly and instead creates a utility
class which is extended with the appropriate sanitization methods.


jackcasey:
Thanks!

I had a quick try of this fix and unfortunately it still doesn't work (in rails 5.1.7 at least) because even via the concern being mixed in that method is private:

```
NoMethodError (private method `sanitize_sql_like' called for Graphiti::Adapters::ActiveRecord::Sanitizer:Class):

/Users/jack/dev/graphiti/lib/graphiti/adapters/active_record.rb:316:in `block in sanitized_like_for'
/Users/jack/dev/graphiti/lib/graphiti/adapters/active_record.rb:314:in `map'
/Users/jack/dev/graphiti/lib/graphiti/adapters/active_record.rb:314:in `sanitized_like_for'
/Users/jack/dev/graphiti/lib/graphiti/adapters/active_record.rb:65:in `filter_string_match'
```

I will help with the fix seeing as you can't repro but I need to head off for today. 👋

wadetandy:
@jackcasey Try now.  Turns out I thought our appraisal suite was testing all of rails 5.0, 5.1, and 5.2 but was only testing against 5.2.  Added appraisals for older rails versions and revealed the problem, which i fixed

jackcasey:
@wadetandy Works great thanks!

jackcasey:
Can probably be closed now btw, as #219 is merged.

PR #209, opened 2020-02-06
Allow resource level links to be defined

superslau:
Graphiti supports links at a relationship level and at the top-level of the
jsonapi document.

Relationship level links are mainly to support lazy loading, and top level links
are used solely for pagination.

This commit adds support for defining links at the resource level.

These links could be used to provide the URL to the resource, or alternative
representation of the resource, or even actions to perform against the resource.
```
{
  "data": {
    "id": 1,
    "type": "tests",
    "attributes": {},
    "relationships": {},
    "links": {
      "self": "http://test.com/api/v1/tests/1"
    }
  }
}
```
To define a link on a resource:
```
link :self do |model|
  url_for([:api, :v1, model])
end
```

### DIFF
--- a/lib/graphiti/resource.rb
+++ b/lib/graphiti/resource.rb
@@ -148,5 +148,16 @@ module Graphiti
       end
       response
     end
+
+    def links?
+      self.class.links.any?
+    end
+
+    def links(model)
+      self.class.links.inject({}) do |memo, (name, blk)| 
+        memo[name] = instance_exec(model, &blk)
+        memo
+      end
+    end
   end
 end
--- a/lib/graphiti/resource/configuration.rb
+++ b/lib/graphiti/resource/configuration.rb
@@ -198,6 +198,7 @@ module Graphiti
               extra_attributes: {},
               sideloads: {},
               callbacks: {},
+              links: {},
             }
         end
 
@@ -235,6 +236,10 @@ module Graphiti
 
         def default_filters
           config[:default_filters]
+        end
+
+        def links
+          config[:links]
         end
       end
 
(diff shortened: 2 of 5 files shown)


superslau:
I've not got any tests - if this PR is going in the right direction I can flesh it out and add some tests

richmolj:
This is really wonderful, thanks ❤️ 

All looks good great (aside from tests)! My only thought here is

```ruby
url_for([:api, :v1, model])
```

Is fine, but might be nice if we could leverage `Resource#endpoints` / `primary_endpoint`. IOW we already have the concept of "this resource is associated to this endpoint" so I wonder if there's a better way to generate that link. Maybe not this specifically but

```ruby
"#{self.endpoint[:url]}/#{model.id}"
```

Should do it. Completely optional, just something to think about.

superslau:
The code example showing an example link to self block was purely for demonstration purposes - but I guess do you want the `self` link to be part of Graphiti? I know jsonapi-resources implements it - maybe that might be a future PR.

In any case, I've updated the PR description and commit message to use `self.endpoint` which is much nicer.

I've submitted a fixup that fixes the tests and adds some tests to cover resource level links.

superslau:
@richmolj I made a small change to my last commit to ensure nil links are included, to make sure spraypaint updates the local instance of the model correctly. 

If that all makes sense then I'll rebase to get this ready for merge

jackcasey:
@richmolj @superslau How is this one going? I'm happy to update this w/ master, if you don't have time.

richmolj:
@jackcasey would appreciate the help - I'm good if we update with master

superslau:
@richmolj @jackcasey I'll try and get this up to date

jackcasey:
FYI I did a PR into this: superslau/graphiti#1 to bring in master.

Also can put it up as a PR into the main repo if that is easier.

PR #210, opened 2020-02-10
Fixed selecting serializer for deeply inherited resources

piotrekbator:
For models `A`, `B` and `C`, where `C` inherits from `B` and `B` inherits from `A` graphiti selected a `BResource` serializer when serializing `C` model records when it should have selected a `CResource`.

This happened because `resource_for_model` method in `polymorhphism.rb` resource class was matched with the model by comparing model class using `is_a?` method which returns `true` when the compared class it's either the class itself or is anywhere in its inheritance chain.

I've changed this comparison to using `==` but also kept the original comparison as a fallback for when nothing is matched using the new method - so that it doesn't break for users that depended on the comparison working as it used to.

I hope that adding a new class to PORO (`PORO::GoldVisa` and a resource for it) isn't an overkill.

### DIFF
--- a/lib/graphiti/resource/polymorphism.rb
+++ b/lib/graphiti/resource/polymorphism.rb
@@ -67,7 +67,8 @@ module Graphiti
         end
 
         def resource_for_model(model)
-          resource = children.find { |c| model.is_a?(c.model) }
+          resource = children.find { |c| model.class == c.model } ||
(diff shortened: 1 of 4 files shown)


richmolj:
Works for me! Thanks @piotrekbator and sorry for the delay ❤️ !

PR #214, opened 2020-03-05
Include more information in RecordNotFound errors.

jackcasey:
When supplying a relationship to an existing resource, if the record cannot
be found then supply the resource name, id and relative json path in the
`Graphiti::Errors::NotFound` exception raised.

### DIFF
--- a/lib/graphiti/errors.rb
+++ b/lib/graphiti/errors.rb
@@ -722,6 +722,21 @@ module Graphiti
     end
 
     class RecordNotFound < Base
+      def initialize(resource = nil, id = nil, path = nil)
+        @resource = resource
+        @id = id
+        @path = path
+      end
+
+      def message
+        if !@resource.nil? && !@id.nil?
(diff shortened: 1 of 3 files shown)


jackcasey:
What do people think of a change like this? 

Putting it up as a RFC / WIP.

aribouius:
I was about to open an issue on this topic, having some way to identify or modify this behavior would be really nice.

Including additional information in the error payload seems like a good first step in improving how the client can handle this.

I'm also curious whether there is some way to handle this scenario without throwing, and instead allowing model validation errors to deliver the message.

richmolj:
I agree this is great, thanks @jackcasey ! If you can please make the standardrb check pass I will happily merge.

To @aribouius there may be additional improvements we can make, but let's tackle them in a separate issue

aribouius:
@richmolj yep totally agree 👍

jackcasey:
@richmolj ✅

richmolj:
Thanks so much @jackcasey ❤️ !

PR #215, opened 2020-03-05
Make context accessible when customizing relationship params

m1foley:
`context` can be useful when customizing params in a relationship. This makes it accessible in the `params` block by adding it as the third block variable:

```rb
class PostResource < ApplicationResource
  has_many :comments do
    params do |hash, _posts, context|
      hash[:filter] = { locale: context.locale }
    end
  end
end
```

The current workaround is to directly reference `Graphiti.context[:object]` within the block, which is less desirable because that private API could change.

### DIFF
--- a/lib/graphiti/sideload.rb
+++ b/lib/graphiti/sideload.rb
@@ -197,7 +197,7 @@ module Graphiti
 
       with_error_handling Errors::SideloadParamsError do
         params = load_params(parents, query)
-        params_proc&.call(params, parents)
+        params_proc&.call(params, parents, context)
         return [] if blank_query?(params)
         opts = load_options(parents, query)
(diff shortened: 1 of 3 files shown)


richmolj:
Hey @m1foley thanks for this ❤️ ! Sorry for the delay, I was traveling then coronavirus :) I appreciate the comments you left! I think this is all good to go if you can fix the conflicts and get CI passing. Thanks!

m1foley:
> I think this is all good to go if you can fix the conflicts and get CI passing

Done! It looks like CI was failing because of an unrelated intermittent issue. Merge conflict fixed.

richmolj:
Much appreciated @m1foley sorry for the delay ❤️ !

PR #217, opened 2020-03-10
Add option to deny empty filter values.

jackcasey:
`deny_empty` is an option that can be set on a filter to raise an error
(a Graphiti::Errors::InvalidFilterValue error) if an empty value is
passed as the filter value.

It can be set at the resource level as a default and/or on individual
filters.

Default behaviour is setting this to false, which will not do anything
special with 'empty' filter values.

Setting on a Resource:
```
class BaseResource < Graphiti::Resource
  self.filters_deny_empty_by_default = true
end
```

Setting on an individual filter:
```
class UserResource < BaseResource
  filter :name, deny_empty: true
end
```

### DIFF
--- a/lib/graphiti/errors.rb
+++ b/lib/graphiti/errors.rb
@@ -143,8 +143,13 @@ module Graphiti
       def message
         allow = @filter.values[0][:allow]
         deny = @filter.values[0][:deny]
+        value_string = if @value == "(empty)"
+          "empty value"
+        else
+          "value #{@value.inspect}"
+        end
         msg = <<-MSG
-          #{@resource.class.name}: tried to filter on #{@filter.keys[0].inspect}, but passed invalid value #{@value.inspect}.
+          #{@resource.class.name}: tried to filter on #{@filter.keys[0].inspect}, but passed invalid #{value_string}.
         MSG
         msg << "\nAllowlist: #{allow.inspect}" if allow
         msg << "\nDenylist: #{deny.inspect}" if deny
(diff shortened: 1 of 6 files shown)


jackcasey:
Requesting comments on this new filter option.

We wanted to deny empty filter values explicitly rather than ignore them or pass them along which seems to be the currently supported behaviour. 

Here is my attempt at implementing this elegantly within graphiti. Another option might be to make the existing `allow_nil` option into a tri-state option; maintaining existing behaviour. I.e: (false = ignore, true = pass, :raise = raise error) but that seemed kind of gross so I went with a new option which overlaps a bit with `allow_nil` but is explicitly about rejecting empty filter values. This seems to work well for us.

richmolj:
Thanks @jackcasey, loving the ability to do this at a resource or filter level. My question: since `allow_nil` allows the string `"null"` (rather than empty value), should we just never allow this? What's the use case we would want to allow it?

jackcasey:
I'm not sure I'm understanding your question correctly.

But I guess there are a few use cases for wanting to get 'empty' values through. Something like `?filter[middlename]=` (to find records where middlename is, actually empty) might be preferable to consumers to `?filter[middlename]=null`.

Our use case is that we want to be explicit about when filters are 'empty' and pre-emptively error out. This was primarily to match our API's existing behaviour as we're doing a refactor (from jsonapi_resources) and don't want to make a version breaking change.

There is definitely some overlap between `allow_nil` and `deny_empty` as well as some discrepancies and could be consolidated and tidied up. Also `allow_nil` was raising some 500's for us. (Something about visiting a NilClass Arel Node or something off the top of my head) I can try to reproduce some of those as failing tests if you like.

richmolj:
My general thought is this should be the default behavior, and you should explicitly pass `null` if you want to find something actually empty. But seems find to merge this for now either way - good to go if you can get CI passing. Thanks 👍 !!

jackcasey:
@richmolj ✅

richmolj:
You're awesome 💯 👏 !!

PR #219, opened 2020-03-19
Hotfix/207

jackcasey:
Addresses #207 
Properly use rails built in sanitisation for doing `LIKE` matches.

I took @wadetandy's PR #208 and rebased off master and made some more changes to the appraisal set up.

Sorry if this is an inappropriate way to go about things. I am trying to get all the work our app relies on into upstream for obvious reasons :)


richmolj (approved):
Solid, thank you ❤️ ! More than happy to upstream 👍

PR #224, opened 2020-04-21
Ws/min 2309 disable links on demand through url query

wilsonsilva:
Pagination links have to be processed and they perform `count` queries. We found out that removing these pagination links greatly improved the performance of our system. Mostly because it is very expensive to do counts on certain tables.

This change allows the API clients to remove the links and pagination links with a query parameter:

```json
api.example.com/resource?links=false&pagination_links=false
```

### DIFF
--- a/lib/graphiti/query.rb
+++ b/lib/graphiti/query.rb
@@ -24,6 +24,8 @@ module Graphiti
 
     def links?
       return false if [:json, :xml, "json", "xml"].include?(params[:format])
+      return false if [false, "false"].include?(@params[:links])
+
       if Graphiti.config.links_on_demand
         [true, "true"].include?(@params[:links])
       else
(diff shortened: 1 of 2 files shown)


wilsonsilva:
If we enable them, they will be enabled for all resources and all requests. Our use case is to configure Graphiti to enable them for all requests and disable them on demand, via a query parameter, in requests that we know we don't need pagination.

wilsonsilva:
We enabled them in the initializer for all requests: 

```ruby
Graphiti.configure do |config|
  config.pagination_links = true
end
```

In some requests for a resource, we want the pagination. In other requests for the same resource, we don't want it. So we disable it via the query parameter.

richmolj:
I'm down with something like this but think the global config should match what we do with `links_on_demand` (in other word "globally enable" and "globally enable but require querystring opt-in" are two separate config options.

PR #230, opened 2020-05-17
Fix included extra fields

### LINKED ISSUE #229: **on_extra_attribute** block is not executed when resource is side loaded
## Background

We have a resource with an `extra_attribute` that sources data from a separate database query, and we depend on `on_extra_attribute` to update the resource state as a cue to execute the query on as "as needed" basis. 

We've noticed that when this resource is "side-loaded" from another resource, the `on_extra_attribute` block is not executed.

Please see this [Slack thread](https://graphiti-api.slack.com/archives/C5A4UEMGS/p1547479536141600) for another user report.

## Expected Results
`on_extra_attribute` block runs regardless of context if a client requests the extra attribute on the resource

## Actual Results
`on_extra_attribute` block is not executed when resource is side-loaded.

gfmurphy:
## Why?
This is an attempt to resolve #229 by including the `extra_attributes` in an association query that are relevant to the resource. This change causes the `on_extra_attributes` hooks to fire for the associated resources, which is what we need to happen.

I'm mostly unfamiliar with this project's code, and I'm not sure this is the best approach. While this fixes the problem for us, I would like feedback on a strategy for writing a better test that captures the desired behavior: When `extra_attributes` are requests for a side-loaded, associated resource, the `on_extra_attributue` hooks are fired for the associated resource.

### DIFF
--- a/lib/graphiti/query.rb
+++ b/lib/graphiti/query.rb
@@ -47,7 +47,11 @@ module Graphiti
         h[:filter] = filters
         h[:sort] = sorts
         h[:page] = pagination
-        unless association?
+        if association?
+          h[:extra_fields] = {
+             @resource.class.type => extra_fields[@resource.class.type]
+           } if extra_fields.key?(@resource.class.type)
+        else
           h[:fields] = fields
           h[:extra_fields] = extra_fields
         end
--- a/lib/graphiti/query.rb
+++ b/lib/graphiti/query.rb
@@ -49,9 +49,7 @@ module Graphiti
         h[:page] = pagination
         if association?
           resource_type = @resource.class.type
-          h[:extra_fields] = {
-             resource_type => extra_fields[resource_type]
-           } if extra_fields.key?(resource_type)
+          h[:extra_fields] = {resource_type => extra_fields[resource_type]} if extra_fields.key?(resource_type)
         else
           h[:fields] = fields
           h[:extra_fields] = extra_fields
--- a/lib/graphiti/query.rb
+++ b/lib/graphiti/query.rb
@@ -44,16 +44,16 @@ module Graphiti
 
     def hash
       @hash ||= {}.tap do |h|
-        h[:filter] = filters unless filters.empty?
-        h[:sort] = sorts unless sorts.empty?
-        h[:page] = pagination unless pagination.empty?
+        h[:filter] = filters
+        h[:sort] = sorts
+        h[:page] = pagination
         unless association?
-          h[:fields] = fields unless fields.empty?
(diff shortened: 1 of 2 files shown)


richmolj:
Thanks for the great description @gfmurphy! So I actually think you've done this correctly, even testing, though I owe you an explanation.

`query.rb` is the worst part of the codebase, partly for normal legacy code reasons and partly because we're trying to support multiple ways to reference relationships. For something like filtering, the correct way is the relationship name `?filter[employee.special_positions][title][eq]=abc`. But the JSON:API spec also supports the resource "type" as well, so `?filter[positions][title][eq]=abc`, and that should apply to any level of the graph. Finally there's a legacy `?filter[employee][positions]` syntax as well for backwards-compatibility. So all this creates a lot of gunk (though a straight refactor would address a lot).

What `query.rb` tries to do is parse the query string in all these ways and create a nice, normalized, nested hash. The rest of the code can simply walk the nested levels of that hash, passing in the parameters to `AppropriateCorrespondingResource.all(the_nested_params)` - and we've already got a bunch of integration specs working at the individual resource level.

In this case, though, we were only supporting the "type" way - because that's the way the JSON:API spec defines for `fields`, we matched it for `extra_fields`. IIRC, we never added the other flavors for `fields/extra_fields` because it wasn't in the spec and wasn't a priority. This had an adverse effect, though, in that `on_extra_attribute` wasn't called. We were walking the hash and passing in the nested parameters - but these were never nested! They were only top-level. This hasn't been an issue before because `on_extra_attribute` is the only part of the codebase that would care about this, otherwise it's just a serialization issue.

And you have solved this! I think you've not only fixed the issue, but supported the relationship syntax. And this seems correct to me - `query.rb` should be returning a nested hash for this just like everything else, so everything follows the same codepath. This is also why I'm not sure we need another test - as long as that hash is correctly nested, we end up in `PositionResource.all(extra_fields: [:foo])` which is already independently integration tested. You could scan through the other tests for `query.rb` which go through the other syntaxes, but what you've already done is what solves the issue for real users.

If you *did* want to add even more testing past that, I typically do this [as a high level integration test, with Rails enabled](https://github.com/graphiti-api/graphiti/tree/a84285b3a0974d2eecc194b95416fb3a2ff8738f/spec/integration/rails). But I'm not sure it's needed in this case for the reasons above - we've fixed the underlying issue which was in query.rb`.

If that all sounds good to you I'm happy to merge!

gfmurphy:
Thanks @richmolj. I'm 🆒 if you are. Thanks for the context. very helpful 👍

PR #231, opened 2020-05-17
Use Query#params in pagination links instead of Query#hash.

gfmurphy:
## Why?
Using `Query#hash` doesn't correctly render links with an include directive and filtering included relationships. Passing the valid parameters received on the original request seems like ok approach.

### DIFF
--- a/lib/graphiti/delegates/pagination.rb
+++ b/lib/graphiti/delegates/pagination.rb
@@ -20,13 +20,17 @@ module Graphiti
 
       private
 
+      def pagination_params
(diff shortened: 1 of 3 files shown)


gfmurphy:
cc: @m1foley

richmolj:
This looks great @gfmurphy I appreciate the fix ❤️ I had a thought we might want to make that rails-specific array of parameters something configured by `graphiti-rails`, but doesn't hurt anything so I lean towards less hoop-jumping. Thanks!

PR #232, opened 2020-05-20
Resource level links update

jackcasey:
NB: This is a refresh and resubmit of #209 which appears to be abandoned by @superslau

### Original PR comments:

Graphiti supports links at a relationship level and at the top-level of the
jsonapi document.

Relationship level links are mainly to support lazy loading, and top level links
are used solely for pagination.

This commit adds support for defining links at the resource level.

These links could be used to provide the URL to the resource, or alternative
representation of the resource, or even actions to perform against the resource.

```
{
  "data": {
    "id": 1,
    "type": "tests",
    "attributes": {},
    "relationships": {},
    "links": {
      "self": "http://test.com/api/v1/tests/1"
    }
  }
}
```

To define a link on a resource:
```
link :self do |model|
  url_for([:api, :v1, model])
end
```


richmolj:
This is great, thanks for the followup @jackcasey ❤️ !

superslau:
Thanks @jackcasey for getting this through the door 👍

PR #236, opened 2020-05-25
Fix conditional rendering of an extra_attribute (:readable)

michal-kazmierczak:
Both `attribute` and `extra_attribute` methods accept `:readable` option allowing to control rendering of an attribute when a condition is met. 

The behavior is not the same for them though - for `attribute` guard is called via `instance_exec` https://github.com/jsonapi-rb/jsonapi-serializable/blob/v0.3.1/lib/jsonapi/serializable/resource/conditional_fields.rb#L117 while for `extra_attribute` it's called via `instance_eval` https://github.com/graphiti-api/graphiti/blob/v1.2.19/lib/graphiti/extensions/extra_attribute.rb#L46

The latter causes:
```
ArgumentError:
  wrong number of arguments (given 1, expected 0)
# ./lib/graphiti/util/serializer_attributes.rb:59:in `block in guard'
# ./lib/graphiti/extensions/extra_attribute.rb:46:in `instance_eval'
# ./lib/graphiti/extensions/extra_attribute.rb:46:in `block in extra_attribute'
```
(see regression spec)

### DIFF
--- a/lib/graphiti/extensions/extra_attribute.rb
+++ b/lib/graphiti/extensions/extra_attribute.rb
@@ -43,7 +43,7 @@ module Graphiti
         def extra_attribute(name, options = {}, &blk)
           allow_field = proc {
             if options[:if]
-              next false unless instance_eval(&options[:if])
+              next false unless instance_exec(&options[:if])
             end
 
             @extra_fields &&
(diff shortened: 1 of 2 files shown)


michal-kazmierczak:
Hello @richmolj,
could you let me know if this PR is ok to be merged?

richmolj:
@michal-kazmierczak my deepest apologies for letting this sit for so long. I don't know how I missed it. But your PR is perfect, a small change with excellent tests. I am truly sorry for not merging this earlier ❤️

richmolj:
Released in `v1.2.24` which should be building in CI now

PR #240, opened 2020-05-29
Sideload macro bug

seocahill:
Relevant slack discussion here:

https://graphiti-api.slack.com/archives/C5A4UEMGS/p1586786246081700

The example given from  ^ is 

```ruby
class EmployeeResource < Graphiti::Resource
  belongs_to :company, writeable: :user_can_modify_company
  def user_can_modify_company
     CompanyPolicy.new(resource, context.current_user).update?
  end
end
```

I'm assuming from above the intention is to have sideload  configurable per request but currently the `user_can_modify_company` method above instead of being lazily called will be fired once on class load and the return value used thereafter.

I've added a failing spec here which appears to recreate this issue but I'm not sure honestly. 

I did dig around @richmolj for a fix but no joy.  I guess jsonrb still does the rendering so is it the case that graphiti simply passes the include parameter to jsonrb's render method based on resource logic (i.e is this definitely a graphiti issue?)

### DIFF
--- a/lib/graphiti.rb
+++ b/lib/graphiti.rb
@@ -33,7 +33,7 @@ module Graphiti
     resources.each do |resource_class|
       resource_class.sideloads.values.each do |sideload|
         flag = sideload.instance_variable_get(:@readable)
-        next if evaluated = sideload.send(:evaluate_flag, flag)
+        next if (evaluated = sideload.send(:evaluate_flag, flag))
 
         resource_class.serializer.relationship(sideload.name, if: -> { evaluated })
       end
--- a/lib/graphiti.rb
+++ b/lib/graphiti.rb
@@ -29,6 +29,14 @@ module Graphiti
   def self.with_context(obj, namespace = nil)
     prior = context
     self.context = {object: obj, namespace: namespace}
+    # conditional reads per request
+    resources.each do |resource_class|
+      resource_class.sideloads.values.each do |sideload|
+        flag = sideload.instance_variable_get(:@readable)
+        evaluated = sideload.send(:evaluate_flag, flag)
+        resource_class.serializer.relationship(sideload.name, if: -> { evaluated })
+      end
+    end
     yield
   ensure
     self.context = prior
--- a/lib/graphiti.rb
+++ b/lib/graphiti.rb
@@ -33,7 +33,8 @@ module Graphiti
     resources.each do |resource_class|
       resource_class.sideloads.values.each do |sideload|
         flag = sideload.instance_variable_get(:@readable)
-        evaluated = sideload.send(:evaluate_flag, flag)
+        next if evaluated = sideload.send(:evaluate_flag, flag)
+
         resource_class.serializer.relationship(sideload.name, if: -> { evaluated })
       end
     end
--- a/lib/graphiti/resource/dsl.rb
+++ b/lib/graphiti/resource/dsl.rb
@@ -180,7 +180,7 @@ module Graphiti
             options[name] ||= send(:"relationships_#{name}_by_default")
           end
         end
-        private :attribute_option
+        private :relationship_option
       end
     end
   end
--- a/lib/graphiti/sideload.rb
+++ b/lib/graphiti/sideload.rb
@@ -99,7 +99,7 @@ module Graphiti
     end
 
     def readable?
-      !!@readable
+      !!evaluate_flag(@readable)
     end
 
     def writable?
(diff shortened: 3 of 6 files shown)


richmolj:
Thanks for digging in @seocahill ❤️ I do think it's a graphiti issue, though it sure would be nice it jsonapi-rb would handle it :)

Seems like the issue is [here](https://github.com/graphiti-api/graphiti/blob/8ce528c6a5034f862f968da1d8451dbfe2bf911f/lib/graphiti/sideload.rb#L29-L30), where we [fire the method or proc](https://github.com/graphiti-api/graphiti/blob/8ce528c6a5034f862f968da1d8451dbfe2bf911f/lib/graphiti/sideload.rb#L425-L436) when the sideload is instantiated (when you call `has_many`).

I think part of the issue here is in development mode Rails reloads the classes and you won't see a problem. If you fiddled with the linked code, can you get your test to pass?

lib/graphiti.rb
```diff
@@ -29,6 +29,14 @@ def self.context=(val)
...
     prior = context
     self.context = {object: obj, namespace: namespace}
+    # conditional reads per request
+    resources.each do |resource_class|
+      resource_class.sideloads.values.each do |sideload|
+        flag = sideload.instance_variable_get(:@readable)
+        evaluated = sideload.send(:evaluate_flag, flag)
+        resource_class.serializer.relationship(sideload.name, if: -> { evaluated })
+      end
+    end
```
seocahill, line 40: After a lot of digging https://github.com/graphiti-api/graphiti/pull/187 put me on the right track...the above may be ugly but I can confirm it is a working solution (for reads)...so that's a step in the right direction at least.

The issue is that all resource serializer logic pertaining to sideloads is evaluated on class load and the request context has no affect thereafter. 

Interestingly this isn't the case for attributes so perhaps the germ of a tidy solution lies in how they are handled?

I wasn't able to write a test to catch this unfortunately.  To recreate this issue run a production server (cache_classes = true) and add a binding or logger in the conditional method governing the sideload read.

You will observe that it fires on class load only.

```ruby
class EmployeeResource < Graphiti::Resource
  has_many positions, readable: :admin?
end

class PositionResource < Graphiti::Resource
  def admin?
     Rails.logger.debug("readable? guard called")
     context.current_user.admin?
  end
end
```

seocahill:
@richmolj ~I see the spec is broken after applying my patch (it passed before). IRL I see the opposite behaviour, with the patch the sideload macro works (per request,  classes cached) without it, it doesn't~

Spec fixed, forgot to update `readable?`

chris-ritsen:
I tried this out and found that the sideloaded resources were removed from the `relationships` property (correct) but still appear in the `included` array (wrong).

seocahill:
Thanks @chris-ritsen I'll have another look

lib/graphiti/resource/dsl.rb
```diff
@@ -180,7 +180,7 @@ def relationship_option(options, name)
             options[name] ||= send(:"relationships_#{name}_by_default")
           end
         end
-        private :attribute_option
+        private :relationship_option
```
seocahill, line 183: this is an unrelated typo

elDub:
@richmolj: Is there any chance that this fix can be merged in? The bug (introduced in https://github.com/graphiti-api/graphiti/pull/178) has caused us to remain on a old version of Graphiti for quite some time now, and there are many other enhancements and fixes later on that we could really use.

richmolj:
@elDub I'd be up for merging if someone would take ownership and address @chris-ritsen 's issue at minimum. Or maybe we could get a failing test so I could address myself.

elDub:
https://graphiti-api.slack.com/archives/C5A4UEMGS/p1588113083203600?thread_ts=1586786246.081700&cid=C5A4UEMGS

@richmolj: The original committer, @caseyprovost, acknowledged that the guards were errantly being executed at class loading time by the [code that he submitted and got approved](https://github.com/graphiti-api/graphiti/commit/750b7a551aa3e544caeb61ed150e95b431f4d1c9#diff-dfce4d9e5946f456ebf1d2dc40b31e9cR29-R30). If the guards were supposed to act like the other guards, in the context of a model instance, it is pretty clear that they shouldn't be being run when the class is loaded. @caseyprovost offered to submit a fix for it and you gave a pointer, but in the absence of an _actual_ fix, can this be completely reverted? As written, I don't see anything positive that the original commit adds.

richmolj:
@elDub the code does look incorrect, so I've prevented it from firing when the sideload initializes. I've left the specs `xit` and the helper method there so we can call it in a future PR (released in `1.2.28`).

@elDub or @seocahill if you'd like to call that code and support procs/runtime variables, we would want to do that at serialization runtime instead of `with_context`. As it stands this PR would fire all guards for all sideloads even if the request didn't need them.

PR #241, opened 2020-05-29
Allow attribute guard to receive name

seocahill:
We have a use case whereby our authorization is quite granular, i.e. we tailor payloads based on a user/role per request.

This PR would allow attribute guard methods to optionally receive the name of the current attribute in addition to the current object.

```ruby
attribute :foo, :string, readable: :bar?

def bar?(object, attribute)
   # Check if current user has a permission allowing them to read foo attribute
   authorize_attribute(context.current_user, attribute) #=> false
end
```

I realize not many applications will require this level of granularity but it seems like a relatively minor change in the scheme of things so I thought I'd ask anyway.

### DIFF
--- a/lib/graphiti/util/serializer_attributes.rb
+++ b/lib/graphiti/util/serializer_attributes.rb
@@ -55,13 +55,16 @@ module Graphiti
       def guard
         method_name = @attr[:readable]
         instance = @resource.new
+        attribute = @name.to_s
 
         -> {
           method = instance.method(method_name)
           if method.arity.zero?
             instance.instance_eval(&method_name)
-          else
+          elsif method.arity == 1
             instance.instance_exec(@object, &method)
+          else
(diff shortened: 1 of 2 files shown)


Zooip:
I would definitely use this. When using ActiveModel::Serializer,  I delegate read authorizations to a `can_read_attribute?(attribute_name)` on my associated PunditPolicy. I was wondering how I could do that on Graphiti ...

richmolj:
@seocahill don't we need to address https://github.com/graphiti-api/graphiti/pull/240 before this?

seocahill:
@richmolj #240 is unrelated AFAIK. 

Do we think this is a good idea, should I add documentation?

richmolj:
Sorry for the delay @seocahill this is 💯

richmolj:
Released in `1.2.29`

PR #242, opened 2020-06-02
Bump `jsonapi-renderer` to `~0.2.2`

Samsinite:
Bump `jsonapi-renderer` to `~0.2.2` now that (https://github.com/jsonapi-rb/jsonapi-renderer/pull/36) is fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 
 Features:
-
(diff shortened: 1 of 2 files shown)


Samsinite:
Hmm... this only fails on one rails 6 test (`integrated resources and adapters filtering when a datetime lte works`), any ideas? Could this test failure be unrelated?

richmolj:
Yes it's a heisenbug. Thanks for the update, I really appreciate it 🙏 !

PR #243, opened 2020-06-09
Add action to query

danleyden:
When performing a write activity on a resource and sideloading a
secondary, related resource, e.g. by `POST /resource/:id?include=related`
the namespace points to the original action for both the primary and
included resources.

This is useful because it provides context for the overall request, but
is incomplete and misleading for the action required on the included
resource and becomes problematic with authorization where the caller may
have write permission on the primary but only read access to the
sideload, meaning that the `Graphiti.context[:namespace]` cannot be
used in all cases.

Beacause a query is created for each resource, including those that are
sideloaded or sideposted, this change adds the action to the query
and sets it appropriately to `:read` when the action for that resource
is a read-type and prevents the action being incorrectly cascaded to
included sideloads.

### DIFF
--- a/lib/graphiti/query.rb
+++ b/lib/graphiti/query.rb
@@ -11,8 +11,7 @@ module Graphiti
       @params = @params.deep_symbolize_keys
       @include_param = nested_include || @params[:include]
       @parents = parents
-      @action = action || @params.fetch(:action, Graphiti.context[:namespace]).try(:to_sym)
-      @action = %i[index show].include?(@action) ? :read : @action
+      @action = parse_action(action)
     end
 
     def association?
@@ -97,7 +96,7 @@ module Graphiti
               sl_resource = resource_for_sideload(sideload)
               query_parents = parents + [self]
               sub_hash = sub_hash[:include] if sub_hash.key?(:include)
-              hash[key] = Query.new(sl_resource, @params, key, sub_hash, query_parents, :read)
+              hash[key] = Query.new(sl_resource, @params, key, sub_hash, query_parents, :all)
             else
               handle_missing_sideload(key)
             end
@@ -317,5 +316,18 @@ module Graphiti
         end
       end
     end
+
+    def parse_action(action)
+      action ||= @params.fetch(:action, Graphiti.context[:namespace]).try(:to_sym)
+      case action
+      when :index
+        :all
+      when :show
+        :find
+      else
+        action
+      end
+    end
+
   end
 end
--- a/lib/graphiti/query.rb
+++ b/lib/graphiti/query.rb
@@ -1,8 +1,8 @@
 module Graphiti
   class Query
-    attr_reader :resource, :association_name, :params
+    attr_reader :resource, :association_name, :params, :action
 
-    def initialize(resource, params, association_name = nil, nested_include = nil, parents = [])
+    def initialize(resource, params, association_name = nil, nested_include = nil, parents = [], action = nil)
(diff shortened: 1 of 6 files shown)


richmolj:
I think this is fantastic and definitely a problem we should solve. My only issue with the current PR is hooking into query/params. If someone does nothing but `Resource.all.to_a` or `Resource.find(1)` I think it should still work; honoring that plain `.rb` usage is a good north star.

To that end, I am wondering if we can instead override `.all`, `.update_attributes` etc. This will decouple is from params, and we can distinguish between `.find` and `.all` operations (instead of just :read). Something like

```ruby
def update_attributes
  self.action = :update
  # ... code ...
  ensure
    self.action = nil
end
```

The only problem here is sideloading with call `._all` and writes will call `._find` IIRC. These private-to-the-repo methods already accept options, so I am thinking implementation ends something like

```ruby
def all(*args)
  # ... code ...
  _all(some_options, action: :all)
end

# make sure to pass { action: :all } when sideloading, too
```

If you'd like we can put some higher-level methods in as well

```ruby
def reading?
  current_action.in?([:find, :all])
end
```

How does that sound to you?

danleyden:
@richmolj thanks for the time and thoughts.

Completely agree with the idea that the plain ruby option should still work; I don't believe that this changeset impacts that - it is entirely additive, with the only change to the interface being a new read-only method on the `query` that is ignored unless your resource (or adapter) goes looking for it...

I did consider something similar to that and concluded that it probably wasn't the best option. However, that was based on incomplete and likely incorrect knowledge of the system...

My understanding of the "high level" architecture of Graphiti is:
* `Resource` defines the structure of and holds the model instance(s) for a "thing"
* `Query` defines the requested operation, parameters and details specific to a given resource instance within the confines of a single request
* `Scope` places boundaries (filters, pagination) on the requested entity based on the `Query`, i.e. translates those information held in the `Query` about the boundaries to something that the underlying ORM can understand
* `Adapter` creates the scope and uses it and the query to perform the required operations on behalf of the resource
* `Runner` manages the creation of the query and triggers handling of the resource (through the proxy)

Given this understanding, the idea of "what needs to be done with a given resource" should really be the responsibility of the query rather than the resource, and adding that data in to the resource would more tightly couple the data with the operation which feels a little wrong... But I'm probably missing something, so please help me understand what....

I'm also not a fan of changing the "current action" within a single resource through its lifecycle and relying on that being consistent from the outside... Although I don't *know* that it would be an issue, my past experiences of spending weeks hunting down race conditions suggest that idea could easily cause one, particularly with custom adapters.

With the write operations, there is an additional concept being thrown around (`meta`) to allow sideposting with different methods for the different child resources. This looks remarkably like a `Query`, so not entirely sure on the thinking of making those separate. Ideally, I would look to use the same thing in both places, but wanted to avoid making major/breaking changes.


I was wrestling with the distinction between `all` and `find` and came to the conclusion that it likely wasn't important. For my needs (authorization), they are both the same (`all` is just a `find` with fewer restrictions, handled by the scoping), so I would really like to understand the different usecase. In my first implementation, I had them being different strings and found that, for my case it was simpler to have just one. I'm fine to add those in and separate out `:read` in favour of more specific terms, but know we quickly get in to details like "should it be `:show` and `:index`?" and wonder if identifying the many types of 'read' action actually buys us anything.

richmolj:
Thanks so much for this @danleyden ❤️  Much appreciated!

PR #246, opened 2020-06-25
Use joins/preload over includes for many_to_many

wadetandy:
Ran into a strange edge case with activerecord's include behavior when
doing many_to_many relationship on a self-referential model. Still can't
quite figure out what the includes issue is, but I think it is actually
a bug in activerecord.  This commit switches to use a combination of
joins/preload when loading the assocation, which seems to do the trick.


richmolj:
Yeah, I think I found this recently myself in a non-graphiti scenario. Good change 👍

PR #247, opened 2020-06-25
Allow custom names for many_to_many filters

wadetandy:
When creating a many_to_many relationship, graphiti adds a filter to the
foreign resource which is used when sideloading or provided as a link to
the relationship scope if link generation is enabled.  Sometimes,
however, the id of the field being looked up is an internal value and
might not be something you'd want to expose to the API itself.  This
allows us to pass a `filter_name` option when setting up many_to_many,
which it will use for the lookups.

### DIFF
--- a/lib/graphiti/sideload/many_to_many.rb
+++ b/lib/graphiti/sideload/many_to_many.rb
@@ -1,4 +1,10 @@
 class Graphiti::Sideload::ManyToMany < Graphiti::Sideload::HasMany
+  def initialize(name, opts)
+    @filter_name_opt = opts[:filter_name]
+
+    super(name, opts)
+  end
+
   def type
     :many_to_many
   end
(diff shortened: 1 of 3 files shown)


richmolj:
Code seems OK but I want to check on the overall scenario. M2M aside, let's say you want to look up and expose `Post` by a slug instead of an id. In that case we'd encourage that the `data/id` rendered is the slug value, the route is `/posts/:id` but `id` is the slug value, etc. IOW treating it as "resource identifier" not "id attribute". How is this different?

wadetandy:
This wouldn't be any change from the perspective of identifying a given resource and looking it up by `id` (whatever that resource determines `id` to mean for it). The difference is how that id gets referred to by other resources in their filters and links.  Say I have resources `User` and `Group`, with a m2m relationship for group membership.  The join table has non-standard foreign keys `group_id` & `true_user_id`, for whatever reason. With the default behavior, we'd need to filter groups by `true_user_id` to get all the groups that a given member was part of, whereas now we can rename it to `member_id` or `user_id` or whatever matches our domain.  

It doesn't change the actual value we use to look it up, however. If instead we always wanted to lookup by username (`filter[member_username]=wadetandy`), then we'd need to override the created filter ourselves to join on the users table instead of going straight through the membership table.

richmolj:
Makes sense, thanks for the explanation!

PR #249, opened 2020-06-27
fix#248: Fixing readable method possibility on extra attributes.

### LINKED ISSUE #248: Support readable on extra fields
* Right now when we use something like:

```
  extra_attribute :card_id, :string, readable: :admin?
```

Is not working:

```
  0) serialization extra attributes adds extra attributes to the serializer
     Failure/Error: -> {
```

ArmellaJuan:
Should close #248


richmolj:
LGMT, thanks @ArmellaJuan ❤️ !

PR #251, opened 2020-07-01
Many to many test

mkamensky:
Spec showing failure to filter by a many_to_many association name, when the association uses `source: something_else`

### DIFF
--- a/spec/fixtures/legacy.rb
+++ b/spec/fixtures/legacy.rb
@@ -71,6 +71,11 @@ ActiveRecord::Schema.define(version: 1) do
     t.integer :book_id
   end
 
+  create_table :reviews do |t|
+    t.integer :reviewer_id
+    t.integer :reviewed_id
+  end
+
   create_table :users do |t|
     t.timestamps
   end
@@ -185,16 +190,25 @@ module Legacy
     has_many :tags, through: :taggings
     has_many :readerships
     has_many :readers, through: :readerships, source: :user
+    has_many :reviews, foreign_key: :reviewed_id, inverse_of: :reviewed
+    has_many :reviewers, through: :reviews
   end
 
   class User < ApplicationRecord
     has_many :readerships
     has_many :books, through: :readerships
+    has_many :reviews, foreign_key: :reviewer_id, inverse_of: :reviewer
+    has_many :reviewed, through: :reviews
   end
   
   class Readership < ApplicationRecord
     belongs_to :user
     belongs_to :book
+  end
+
+  class Review < ApplicationRecord
+    belongs_to :reviewer, class_name: 'User'
+    belongs_to :reviewed, class_name: 'Book'
   end
 
   class Tag < ApplicationRecord
@@ -241,6 +255,11 @@ module Legacy
     belongs_to :genre
     many_to_many :tags
     many_to_many :readers, resource: Legacy::UserResource
+    many_to_many :reviewers, resource: Legacy::UserResource
+  end
+
+  class UserResource < ApplicationResource
+    many_to_many :reviewed, resource: Legacy::BookResource
   end
 
   class StateResource < ApplicationResource
--- a/spec/fixtures/legacy.rb
+++ b/spec/fixtures/legacy.rb
@@ -66,6 +66,15 @@ ActiveRecord::Schema.define(version: 1) do
     t.timestamps
   end
 
+  create_table :readerships do |t|
+    t.integer :user_id
+    t.integer :book_id
+  end
+
+  create_table :users do |t|
+    t.timestamps
+  end
+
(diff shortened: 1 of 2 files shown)


richmolj:
Thanks @mkamensky this is super helpful. I know the issue will be something in the introspection [here](https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/adapters/active_record/many_to_many_sideload.rb) think you could take a look?

mkamensky:
> Thanks @mkamensky this is super helpful. I know the issue will be something in the introspection [here](https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/adapters/active_record/many_to_many_sideload.rb) think you could take a look?

Yes, I already found it, but I don't really understand what's going on: the statement
```ruby
many_to_many :readers, resource: Legacy::UserResource
```
inside `BookResource` creates a filter for `UserResource`, rather than `BookResource`. I don't see where the filter for `BookResource` would be created

wadetandy:
Hey @mkamensky, I think this is more an issue of a missing feature than a bug. We currently don't do a ton of introspection on the join model to take into account things like source, though this would be a really nice addition. However, you can get what you're looking for if you rebase your PR on top of the current master branch with my recently merged changes from #252. The issue is that you need to define name of the filter on the `UserResource` and leverage the `inverse_filter` option which is added by my merged PR. If you add this to the `before` block to your test context, the test will pass.

```ruby
Legacy::UserResource.class_eval do
  many_to_many :books,
    resource: Legacy::BookResource,
    inverse_filter: :reader_id
end
```

mkamensky:
@wadetandy thanks, that's how I ended up doing it. The reason I think it is more than an inconvenience is that if you have several many-to-many relations with the same model type, you get into trouble. For instance, suppose each book also has authors and reviewers, both of type User. So you may put
```ruby
many_to_many :readers, resource: Legacy::UserResource
many_to_many :authors, resource: Legacy::UserResource
many_to_many :reviewers, resource: Legacy::UserResource
```
inside `BookResource`, and they will silently overwrite each other, even if the join tables are different. This is how I found this issue, and it took some time to figure out what's going on.

With the proposed solution, I should have, inside `UserResource`, something like:
```ruby
many_to_many :books, resource: Legacy::BookResource, inverse_filter: :reader_id
many_to_many :authored, resource: Legacy::BookResource, inverse_filter: :author_id
many_to_many :reviewed, resource: Legacy::BookResource, inverse_filter: :reviewer_id
```
which works, but I find it strange, since `UserResource` contains the knowledge of how `Book` likes to call its readers, authors and reviewers (but perhaps there is no better way). Anyway, if this is the chosen solution, perhaps there should be some warning in case the filter is overwritten?

mkamensky:
Ok, I think I found a reasonable solution: one should make the name changes already on the joining model. So in the example, the join model `Review` for the reviews may look like this:
```ruby
class Review < ApplicationRecord
  belongs_to :reviewer, class_name: 'User'
  belongs_to :reviewed, class_name: 'Book'
end
```
and the other classes:
```ruby
class User < ApplicationRecord
  has_many :reviews, foreign_key: :reviewer_id, inverse_of: :reviewer
  has_many :reviewed, through: :reviews
end

class Book < ApplicationRecord
  has_many :reviews, foreign_key: :reviewed_id, inverse_of: :reviewed
  has_many :reviewers, through: :reviews
end
```
Then in the resources
```ruby
class BookResource < ApplicationResource
  many_to_many :reviewers, resource: Legacy::UserResource
end

class UserResource < ApplicationResource
  many_to_many :reviewed, resource: Legacy::BookResource
end
```
and things seem to work as expected. I uploaded a modified patch with a passing test.

It is still a bit confusing that the last `many_to_many` statement produces a filter in the `Book` resource, but I guess normally one wouldn't notice it.

So from my point of view, there is no need to modify the code, but maybe document all this somewhere?

mkamensky:
By the way, it appears that the `inverse_filter` functionality can also be attained with the `foreign_key` option of `many_to_many`, no?

wadetandy:
`inverse_filter` works in a few more cases than `foreign_key`, as the foreign key actually has to line up with what the tables look like, whereas you can really override the name for the inverse filter to whatever you want, which is helpful especially if you're applying graphiti to a codebase which can't just easily change column names and such without messing with other code.

Glad you found a solution. If you check out my changes from #255, I actually make it so that your original version can work as well, which allows your original join table to reference `user_id` and `book_id` instead of changing them for each join.  You just need to make sure you use `inverse_of` on the join table associations that have non-standard naming.

mkamensky:
@wadetandy That's great, thanks! I somehow missed #255 , it does seem to solve my problems, will test it when it's merged.

PR #252, opened 2020-07-02
Change filter_name => inverse_filter

wadetandy:
Recent feedback from a user trying to implement the new filter_name
functionality was that they were putting the filter_name override on the
resource at the wrong side of the relationship. Adding inverse to the
argument name matches ActiveRecord's `inverse` argument to a
relationship definition and makes things more clear.

Additionally, I realized that overriding the filter from a link
generation perspective is desireable on has_many sideloads as well, so I
pulled this logic up a level into that sideload class.

### DIFF
--- a/lib/graphiti/sideload.rb
+++ b/lib/graphiti/sideload.rb
@@ -132,6 +132,10 @@ module Graphiti
       end
     end
 
+    def link_filter(parents)
+      base_filter(parents)
+    end
+
     # The parent resource is a remote,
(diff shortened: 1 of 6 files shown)


wadetandy:
@mkamensky Heads up on this PR, as it relates to what you were working through today.  Hopefully the naming change helps make which side of the relationship the option goes on a bit more clear.

richmolj:
I think we need to document this but LGTM - thanks!

PR #254, opened 2020-07-13
Fix: filter_datetime_eq method in ActiveRecord adapter doesn't supprt nil

zailleh:
This PR fixes an error seen in the ActiveRecord adapter when attempting to filter using nil.

Scenario: In a Rails app, with a resource similar to this the below. filtering by `{ reviewedAt: { eq: "null" } }` from Spraypaint causes error: `NoMethodError (undefined method '+' for nil:NilClass)`
```ruby
class UserRequestResource < ApplicationResource

  primary_endpoint "/admin/user_requests"

  attribute :action, :string
  attribute :first_name, :string
  attribute :last_name, :string
  attribute :reviewed_at, :datetime
  attribute :approved, :boolean

  belongs_to :requester, :resource => UserResource
  belongs_to :reviewer, :resource => UserResource

  filter :reviewed_at, :allow_nil => true

end
```

My workaround for now has been to define a custom filter, but it's not ideal and a permanent fix would be better.
```ruby
filter :reviewed_at, :allow_nil => true  do
    eq do |scope, value|
      scope.where(:reviewed_at => value)
    end
    not_eq do |scope, value|
      scope.where.not(:reviewed_at => value)
    end
  end
```

### DIFF
--- a/lib/graphiti/adapters/active_record.rb
+++ b/lib/graphiti/adapters/active_record.rb
@@ -159,7 +159,7 @@ module Graphiti
 
       # Ensure fractional seconds don't matter
       def filter_datetime_eq(scope, attribute, value, is_not: false)
-        ranges = value.map { |v| (v..v + 1.second - 0.00000001) }
+        ranges = value.map { |v| (v..v + 1.second - 0.00000001) unless v.nil? }
         clause = {attribute => ranges}
         is_not ? scope.where.not(clause) : scope.where(clause)
       end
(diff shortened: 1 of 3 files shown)


richmolj:
Thanks for the great PR description @zailleh ❤️  Would you mind adding a test? If we had that this is good to go!

zailleh:
> Thanks for the great PR description @zailleh ❤️ Would you mind adding a test? If we had that this is good to go!

Done!

richmolj:
You rock @zailleh thanks ❤️ !

richmolj:
Released in `1.2.20`

PR #255, opened 2020-07-16
Infer inverse_filter from `inverse_of`

wadetandy:
If an active_record has_many association provides the `inverse_of`
option, we can infer better filter name for the opposite side of a
many_to_many relationship

Fixes #251, though with a slight change needed on the ActiveRecord definition.


richmolj (approved):
Great stuff as always @wadetandy ⭐

richmolj:
Released in `1.2.25`

PR #259, opened 2020-07-26
Patch memory leak in Graphiti::Debugger.

### LINKED ISSUE #258: Memory leak in Graphiti::Debugger
## Background

We've noticed that memory usage in our application continues to grow and never plateaus under load in our application. Profiling has found a memory leak in the `Graphiti::Debugger` class. The `chunks` Array, which is referenced by a class instance variable, is never flushed when the debugger is disabled.

## Memory Growth
new_relic_mem

## Expected Results
The memory usage of the debugger should not continue to grow as the application runs

## Actual Results
Profiling shows that the `add_chunk` function continue to add data to the `chunks` array, and it is never flushed if the debug is disabled. This retains references to objects in memory that are never GC'd.

gfmurphy:
## Why?
This an attempt to fix #258 in the most focused way possible.

The `Debugger` was adding data to the chunks Array via the `ActiveSupport::Notification` handlers, `on_data`, and `on_render`. When the debugger is disabled, this Array is never flushed, and it is not GC'd since it's referenced in a class instance variable.

This change returns early in both event handlers if the debugger is not enabled. We also mark the utility methods in the debugger as private so they cannot be accessed outside of the class (and add data to chunks).

My first approach was to add the guard clause directly to `add_chunks` private method, which also fixes the leak; however, i changed to this strategy to save some CPU and memory building the data structures that aren't needed if debugging is disabled.

### DIFF
--- a/lib/graphiti/debugger.rb
+++ b/lib/graphiti/debugger.rb
@@ -10,6 +10,8 @@ module Graphiti
 
     class << self
       def on_data(name, start, stop, id, payload)
+        return [] unless enabled
+
         took = ((stop - start) * 1000.0).round(2)
         params = scrub_params(payload[:params])
 
@@ -24,7 +26,7 @@ module Graphiti
         end
       end
 
-      def on_data_exception(payload, params)
+      private def on_data_exception(payload, params)
         unless payload[:exception_object].instance_variable_get(:@__graphiti_debug)
           add_chunk do |logs, json|
             logs << ["\n=== Graphiti Debug ERROR", :red, true]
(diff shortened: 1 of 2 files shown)


lib/graphiti/debugger.rb
```diff
@@ -24,7 +26,7 @@ def on_data(name, start, stop, id, payload)
         end
       end
 
-      def on_data_exception(payload, params)
+      private def on_data_exception(payload, params)
```
gfmurphy, line 29: I think these methods should be private since they're only used to support the main notification callbacks. This also saves some testing 😄 

I opted for "inline" private modifier to keep diff as small as possible. If preferred I can move this methods under the private group.

richmolj:
This looks good to me for now, and thanks @gfmurphy for the help ❤️  I'm not sure why my own apps haven't seen the same behavior.

richmolj:
Released in `v1.2.21` which CI should be building right now

gfmurphy:
Thanks so much @richmolj! Does that CI job push the gem to rubygems?

richmolj:
Should have but apparently did not...it's now there :)

PR #265, opened 2020-08-18
Better handling of validations

### LINKED ISSUE #264: Primary key needs to be writable in order to perform updates
Hi there,

it seems that graphiti currently requires a primary key to be writable in order to perform any updates. 

**How to reproduce**
https://gist.github.com/apauly/1991e741fcec159632ef6e34866909c9

**What did I expect**
Being able to perform updates without allowing the user to write the primary key which will also allow them to choose a primary key on create.

**Discussion**
It seems that the `Graphiti::Deserializer` is merging the id into the attributes hash:
https://github.com/graphiti-api/graphiti/blob/e63586d26462896880f6299628266561fc9ffa80/lib/graphiti/deserializer.rb#L70

I tried to remove it and the only spec that fails afterwards is the one that explicitly tests this behaviour:
https://github.com/graphiti-api/graphiti/blob/e63586d26462896880f6299628266561fc9ffa80/spec/deserializer_spec.rb#L30

I could not find any reason why this is required at all. Any hints on this?

wagenet:
Ensures that non-writable ids work correctly in assocations and updates
among other things.

Fixes #264

### DIFF
--- a/lib/graphiti/deserializer.rb
+++ b/lib/graphiti/deserializer.rb
@@ -45,9 +45,27 @@
 #
 #   { type: 'authors', method: :create, temp_id: 'abc123' }
 class Graphiti::Deserializer
-  def initialize(payload = {})
+  CONVERSIONS = {
+    build: :create,
+    find: :update
+  }
+
+  def initialize(payload, action)
     @payload = payload
     @payload = @payload[:_jsonapi] if @payload.key?(:_jsonapi)
+    @action = action.to_sym
+  end
+
+  # We may start off with a build or find payload that then we later learn
+  # are actually create or update. This allows us to make that transition.
+  def convert(new_action)
+    return if new_action == @action
+
+    unless CONVERSIONS[@action] == new_action
+      raise ArgumentError, "can't convert #{@action} to #{new_action}"
+    end
+
+    @action = new_action
   end
 
   def params
@@ -64,11 +82,14 @@ class Graphiti::Deserializer
     data[:id]
   end
 
-  # @return [Hash] the raw :attributes hash + +id+
+  # @return [Hash] the raw :attributes hash + +id+ if creating
   def attributes
-    @attributes ||= raw_attributes.tap do |hash|
-      hash[:id] = id if id
-    end
+    @attributes ||=
+      if id && include_id_in_attributes?
+        raw_attributes.merge(id: id)
+      else
+        raw_attributes
+      end
   end
 
   # Override the attributes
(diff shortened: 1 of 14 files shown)


wagenet:
This ended up being a bit more far-reaching than I originally expected. I don't think any public APIs are affected, however, I could see some behaviors changing a bit unexpectedly. I _think_ the new behaviors would be more correct, but this might also be considered breaking.

richmolj (dismissed):
@wagenet you're an absolutely spectacular human being, can't thank you enough for the help here 🙏 

The amount of changes needed are a bit disappointing, though I can see why they were necessary. What do you think about using `Graphiti.context[:namespace]` instead of passing around `action`? Though rule-of-thumb I'd pass it around, I think this is a good exception to the rule and unique to how Graphiti works (everything happens within a given context, the action is part of that context, so make it easily accessible everywhere). We already do this in a couple other places.

One downside I see is it's not as friendly to plain ruby (though maybe we could set the namespace when calling `.build` etc).

This is my main thought to make things a little less cumbersome - I applaud you for wading through these moving parts - but it's a toss-up for me and I trust your judgement, so approving the PR either way.

Again, thanks so much for this and sorry it was this much of a hassle ❤️

richmolj:
> I think the new behaviors would be more correct, but this might also be considered breaking.

I agree with this, though lean more towards "correct" than "breaking" in this case

wagenet:
@richmolj yeah, there's some parts of this I'm not 100% in love with and I would prefer not to make the internals worse. That said the specs are passing (except for some failures in appraisals, which I need to investigate) and it works for us, so that counts for something.

wagenet:
@richmolj I did a bit of refactoring. I also split up the commits to make it clear that there are two distinct issues being solved here. IMO, it's ok to pass around action since it very distinctly affects how we do stuff. The only other potentially better solution I can see is to not have the Deserializer and Validations so deeply embedded.

richmolj:
@wagenet spent time on this this morning, and sorry I wasn't able to spend more. I have a few notes here, though I wasn't able to get something working in the time that I had. Feel free to continue on your current path, just providing this in case it's helpful.

* Instead of merging the `id` in the deserializer, we might want to do this directly in `.build`. There's no other scenario it makes sense to perform this logic, so instead of passing the action around and performing conditional logic, let's just move the logic to the one place it makes sense. I *think* this would avoid all the action/convert stuff in both the deserializer and validator.

* We should probably only run [this line](https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/deserializer.rb#L178) when the `method` (a couple lines below) is `create`. This is similar to the `.build` point above.

* We should probably [not validate in the runner](https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/runner.rb#L14). Not only does this cause the needed `.convert` of the action, we're validating a write payload during a read operation (`.find` will later call `.update_attributes`, but only the latter should run the validator) which seems odd to begin with. Maybe not relevant to this PR but I find it odd.

I think there's more fundamental refactoring that would make things easier but neither of us has time for. The first two bullet-points seem to me the simplest fix without that overhaul, but I haven't had the time to follow through more. Hope this helps.

aribouius:
> > I think the new behaviors would be more correct, but this might also be considered breaking.
> 
> I agree with this, though lean more towards "correct" than "breaking" in this case

I also agree this is more "correct", but I'd suggest it be treated as breaking.  Consider that some apps might be taking advantage of the fact that associations are receiving the callbacks, as a means of permission enforcement.

Would it be feasible to make this configurable, defaulting to current behavior for now?

PR #267, opened 2020-08-22
Avoid need for writable PK on update

### LINKED ISSUE #264: Primary key needs to be writable in order to perform updates

richmolj:
There's some larger refactoring needed but this is the simplest possible
fix for the issue. There are three overall problems with the validator:

* Called for read operations, but meant to validate writes. I'm not sure
why this was ever the case, perhaps to eventually validate reads as
well.
* Called in multiple places (proxy and runner)
* References `params[:action]`, which couples us to Rails

So the ideal refactoring would be to only call the validator within
`ResourceProxy#save`, which addresses all the above.

The problem is one specific test looks at the metadata of a sideposted
resource (occasionally it's helpful to see things like "what parent
object is sideposting me"). Because read operations would go through the
validator and `id` would be typecast, this test now fails because it
sees `id` as a string instead of an integer, because it is no longer
being typecast.

I think I'm fine just updating the spec, but again this PR is to do the
simplest thing possible. So I instead added the very specific logic "if
validating id, and not a create operation, AND id is not writable, don't
attempt to typecast". The larger refactoring can come later.

Finally, the other future refactoring would be to avoid merging `id`
into `attributes` when not a create operation. Part of the issue here is
the deserializer is called even on reads (per above), which complicates
things further. So a good future approach would be to only call the
validator within `ResourceProxy#save`, then address deserialization
separately afterwards.

Fixes https://github.com/graphiti-api/graphiti/issues/264
See also: https://github.com/graphiti-api/graphiti/pull/265

### DIFF
--- a/lib/graphiti/request_validator.rb
+++ b/lib/graphiti/request_validator.rb
@@ -6,18 +6,18 @@ module Graphiti
       :deserialized_payload,
       to: :@validator
 
-    def initialize(root_resource, raw_params)
-      @validator = ValidatorFactory.create(root_resource, raw_params)
+    def initialize(root_resource, raw_params, action)
+      @validator = ValidatorFactory.create(root_resource, raw_params, action)
     end
 
     class ValidatorFactory
-      def self.create(root_resource, raw_params)
-        case raw_params["action"]
-        when "update" then
+      def self.create(root_resource, raw_params, action)
+        case action
+        when :update then
           RequestValidators::UpdateValidator
         else
           RequestValidators::Validator
-        end.new(root_resource, raw_params)
+        end.new(root_resource, raw_params, action)
       end
     end
   end
--- a/lib/graphiti/request_validators/validator.rb
+++ b/lib/graphiti/request_validators/validator.rb
@@ -3,10 +3,11 @@ module Graphiti
     class Validator
       attr_reader :errors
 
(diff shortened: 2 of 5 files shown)


wagenet (approved):

wagenet:
@richmolj what's up with the failing CI?

richmolj:
Heisenbug I haven't tracked down yet 😬

richmolj:
Released in `v1.2.23` which should be building in CI now

PR #274, opened 2020-09-02
Revert change from includes to preload/join

wadetandy:
This was causing some pretty extreme performance problems for a couple
of our users, and it seems I've been the only one to ever encounter this
strange edge case, which is easy enough to override in my specific
resource in the two places I've seen it.


richmolj (approved):

richmolj:
Released in `1.2.27`

PR #278, opened 2020-09-13
fixes ruby 2.7 kwargs usage warnings

mariacorazon:
fixes several warnings when using ruby version 2.7.0. 

a sample warning below:
> graphiti/lib/graphiti/resource/configuration.rb:132: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
>graphiti/lib/graphiti/resource/configuration.rb:253: warning: The called method `get_attr' is defined here

[more information about this ruby 2.7.0 warning can be read here](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/#:~:text=In%20Ruby%202.7%2C%20keyword%20arguments%20can%20use%20non%2DSymbol%20keys.&text=If%20a%20method%20accepts%20both,non%2DSymbol%20keys%20are%20allowed.)

### DIFF
--- a/lib/graphiti/hash_renderer.rb
+++ b/lib/graphiti/hash_renderer.rb
@@ -45,10 +45,10 @@ module Graphiti
       {}.tap do |hash|
         hash[:data] = if serializers.is_a?(Array)
           serializers.map do |s|
-            s.to_hash(opts)
+            s.to_hash(**opts)
           end
         else
-          serializers.to_hash(opts)
+          serializers.to_hash(**opts)
         end
       end
     end
--- a/lib/graphiti/request_validators/validator.rb
+++ b/lib/graphiti/request_validators/validator.rb
@@ -51,7 +51,7 @@ module Graphiti
           relationships: relationships
         }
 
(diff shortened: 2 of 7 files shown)


wadetandy:
This looks great! Only thing I’d like to see added is putting ruby 2.7 in our Travis.yml config to make sure we are keeping this working moving forward. Should be able to go everywhere 2.6 is already referenced.

mariacorazon:
> This looks great! Only thing I’d like to see added is putting ruby 2.7 in our Travis.yml config to make sure we are keeping this working moving forward. Should be able to go everywhere 2.6 is already referenced.

@wadetandy, from travis build result [https://travis-ci.org/github/graphiti-api/graphiti/builds/726864762](https://travis-ci.org/github/graphiti-api/graphiti/builds/726864762) I can see that ruby 2.7 was also referenced. is that what you meant? thanks.

wadetandy:
Seems I did that already and didn't remember doing it! Sorry about that.

PR #279, opened 2020-09-24
adapter support for non sql based databases

mrhardikjoshi:
[slack reference](https://graphiti-api.slack.com/archives/C5A4UEMGS/p1588869748275000)

While trying to adapt Graphiti to neo4j database via its [OGM](https://github.com/neo4jrb/activegraph). We found that activerecord is engraved into Graphiti in some places, and we would need a cleaner separation between graphiti and activerecord to make Graphiti easy to use for other type of databases.

I have tried to separate the AR dependancy from some modules and I have moved it back to adapter classes. It would be much easier to use Graphiti with non relational databases if relationalDB concept such as foreign_key etc are placed in adapters.

I have not added test cases related to the changes to avoid rework in case of some changing feedback, once the initial review of changes are done I can add tests related to the changes.

### DIFF
--- a/lib/graphiti/delegates/pagination.rb
+++ b/lib/graphiti/delegates/pagination.rb
@@ -6,7 +6,7 @@ module Graphiti
       end
 
       def links?
-        @proxy.query.pagination_links?
+        @proxy.query.pagination_links? && @proxy.data.present?
       end
 
       def links
--- a/lib/graphiti/adapters/abstract.rb
+++ b/lib/graphiti/adapters/abstract.rb
@@ -400,6 +400,9 @@ module Graphiti
         raise "you must override #destroy in an adapter subclass"
       end
 
+      def close
+      end
+
       def self.numerical_operators
         [:eq, :not_eq, :gt, :gte, :lt, :lte].freeze
       end
--- a/lib/graphiti/adapters/abstract.rb
+++ b/lib/graphiti/adapters/abstract.rb
@@ -1,6 +1,9 @@
 module Graphiti
   module Adapters
     class Abstract
+      require "graphiti/adapters/persistence/associations.rb"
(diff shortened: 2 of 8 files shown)


lib/graphiti/delegates/pagination.rb
```diff
@@ -6,7 +6,7 @@ def initialize(proxy)
       end
 
       def links?
-        @proxy.query.pagination_links?
+        @proxy.data.present? && @proxy.query.pagination_links?
```
mrhardikjoshi: adding an extra check of `@proxy.data.present?` here. Do we really need to render pagination links if the records on requested endpoint are empty?

richmolj:
@mrhardikjoshi this is great! I support all of these changes if we can break them into separate commits (something like `#close` or pagination links can be separated) with tests when necessary. Thank you so much ❤️ !

A few notes on the general premise. One is that the *terminology* `foreign_key` is specific to relational DBs, but the *concept* is not. I think the concept remains valid and important.

One related aspect there, which I imagine is the cause of the PR to begin with, is dealing with embedded documents. If you cannot access a child document without its parent, if you cannot deal with an entity independently, then this is a better fit for a nested attribute than a relationship. If you *can* deal with entities independently, then some identifier linking entities is required.

All the changes here simply take the existing functionality and move to the adapter, so we're all good. But worth noting the general issue as it often comes up with folks using non-relational DBs.

mrhardikjoshi:
>I support all of these changes if we can break them into separate commits (something like #close or pagination links can be separated) with tests when necessary.

@richmolj Apology for late update, I have divided the logic in different commits as requested. Please review them when you get chance. Thank you.

richmolj:
I think we're good to go! Again, great work on this @mrhardikjoshi 👏 🎉 !

Issue #287, opened 2020-10-09
ActiveRecord::StatementInvalid when query count statistics with a complex select query in base_scope


tranxuanthang:
**Issue:** When getting count statistics, for example, with `stats[total]=count` request, and `base_scope` of the resource has some specific complex `select` query, the count query will fail with `ActiveRecord::StatementInvalid` exception.

The failing `base_scope` usually has some `AS` keyword in it. For example:

```ruby
def base_scope
  User.select("*", "name AS first_name")
end
```

**How to reproduce:** Define a `base_scope` that has a chaining with `select` method like above, and then query the resource with `{ stats: { total: "count" } }`.

**Expected**: The query runs sucessfully, with correct count number, and includes anything on `select` query.

**Actual:** The query fails to run with `ActiveRecord::StatementInvalid` error:

```shell
Failure/Error: scope.distinct.count

     ActiveRecord::StatementInvalid:
       SQLite3::SQLException: near "*": syntax error
```

tranxuanthang:
We can close the issue for now, because the fix has been merged.

PR #288, opened 2020-10-09
Fix total count not working with some base_scope when using Active Record

tranxuanthang:
This is my attempt to fix issue #287.

It may be related to how the Active Record's `count` method changed in Rails version 4. TLDR is we should use `Model.count(:all)` instead of `Model.count`:
- https://github.com/rails/rails/issues/15138
- https://stackoverflow.com/questions/18252570/rails-difference-between-model-count-and-model-countall

P/S: This is one of my first pull requests to any repositories other than my own, so forgive me if I missed anything.

If you think my pull request does help, you may label this pull request as `hacktoberfest-accepted` since I'm participating in hacktoberfest event this month.

The last thing is thanks for creating an awesome library, I've been currently using it in production and it works really well!

### DIFF
--- a/lib/graphiti/adapters/active_record.rb
+++ b/lib/graphiti/adapters/active_record.rb
@@ -191,7 +191,7 @@ module Graphiti
       # (see Adapters::Abstract#count)
       def count(scope, attr)
         if attr.to_sym == :total
-          scope.distinct.count
+          scope.distinct.count(:all)
         else
           scope.distinct.count(attr)
         end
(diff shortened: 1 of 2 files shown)


richmolj:
Looks great @tranxuanthang, I appreciate the spec and links 👍 ! If you wouldn't mind changing the spec description to something like "when select statement is specified", just so we know the context if if ever fails, I'd be good to merge!

tranxuanthang:
Hi @richmolj, thank you for your response! I've just updated the description of the spec as you described. Please feel free to let me know if there are any other problems.

richmolj:
Thank you @tranxuanthang !

PR #295, opened 2020-10-27
Do not generate many_to_many filter when it exists

hanssgo:
- Fixes an issue where we are unable to modify the generated filter
because it gets blown away by the generated filter.


hanssgo:
@richmolj I'm not sure how to fix the failing spec. It passes when run by itself but fails when run as a suite. It looks like the previous test is affecting it.

hanssgo:
@richmolj ready for review & merging

richmolj (approved):

richmolj:
Awesome 🎉 ! Thanks @hanssgo !

PR #297, opened 2020-11-04
Fix logger level check

budu:
Use public method `#debug?` instead of checking the numeric
representation of logger level.

This fix compatibility issue with alternative loggers, see
rocketjob/semantic_logger#28.


richmolj:
Thanks @budu!

Issue #198, opened 2020-01-14
Possible Bug: sideposting via PATCH does not pick up errors on has_many children when only posting new ones


mattgibson:
The scenario: 
I have a model with a has_many relationship, with an instance and some of its children already persisted. I want to add a new one. I send a PATCH request with only the to-be-created child object, skipping the existing ones that I want to leave alone. All works fine unless the new child is invalid. In that case, the error is ignored and a 200 response is returned with no errors.

The problem:
The reason is the [following code](https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/util/validation_response.rb#L61-L63) in `Graphiti::Util::ValidationResponse`:
```
        related_objects.each_with_index do |r, index|
          method = payload[index].try(:[], :meta).try(:[], :method)
          next if [nil, :destroy, :disassociate].include?(method)
```

Because the `index` is from the existing association on the model, it includes all the DB models as well as the one I am adding, so when adding a fourth to three existing ones, the index for the new one is `3`. However, the payload only had the one I am adding, so to find the params, we would need to use `payload[0]`. 

I am unsure if this is actually a bug or if I am misusing the sideposting mechanism. The spec and docs don't seem to mention whether a PATCH can be done this way.

richmolj:
Hey @mattgibson ! 👋 

> The spec and docs don't seem to mention whether a PATCH can be done this way.

You should definitely be able to do this. I think the big thing is

> Because the index is from the existing association on the model, it includes all the DB models as well as the one I am adding

We actually run some [wonky code](https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/adapters/active_record.rb#L228-L239) to make sure `model.relationships` only returns the records sent in the request. So we might have a bug, unless you're on an older version of Graphiti or running some custom code in your app? It would be a big help to see this reproduced in either the sample app or a test within Graphiti.

mattgibson:
Ah, OK. Sounds like a bug, then. I stepped throught that code with the debugger and `association` already contains the existing 3 records at that point, so it adds the to-be-created one.

Maybe a solution is to loop over the objects in the relationship payload matching by `id` or `temp-id`, but I don't know if `temp-id` is still available at that point.

I'll have a go at making a test for it

richmolj:
It should be available as a magic variable on the model instance, e.g. `@post.instance_variable_get(:'@_jsonapi_temp_id')` (unavoidable 😝 )

> already contains the existing 3 records at that point

[...]

PR #304, opened 2020-12-09
Fix: Uncaught ActiveRecord::RecordInvalid error when saving new, invalid has_many or many_to_many associations

zailleh:
I discovered that validation errors were not coming through to the frontend when saving an invalid association on a `many_to_many` join; similar to (or the same as, I'm not sure): https://github.com/graphiti-api/graphiti/issues/198

The root cause was `parent.send(association_name) << child` in the ActiveRecord adaptor since this causes a `save!` operation which throws the `ActiveRecord::RecordInvalid` error, which is uncaught, causing the JSON API response to not contain relevant information as to what the validation error was; or in our use case, the error was caught by a global error handler and returned a non JSON API response.

To resolve the issue and allow validation errors to be returned as expected, create a new association and re-apply instance variables (`@_jsonapi_temp_id` at least is required for generating errors in [GraphitiErrors::Validation::Serializer](https://github.com/graphiti-api/graphiti_errors/blob/4e5341928b7fd39b26aacb544d833080f0f03441/lib/graphiti_errors/validation/serializer.rb#L119)

Simply rescuing the error doesn't work as the new association isn't maintained.

Note: Force push to squash troubleshooting commits into one neat commit

### DIFF
--- a/lib/graphiti/adapters/active_record.rb
+++ b/lib/graphiti/adapters/active_record.rb
@@ -241,7 +241,16 @@ module Graphiti
             if association_type == :many_to_many &&
                 [:create, :update].include?(Graphiti.context[:namespace]) &&
                 !parent.send(association_name).exists?(child.id)
-              parent.send(association_name) << child
+              target = parent.send(association_name)
+              if child.persisted?
+                target << child
+              else
+                new_child = target.create(child.attributes)
+
+                child.instance_variables.each do |variable|
+                  new_child.instance_variable_set(variable, child.instance_variable_get(variable))
+                end
+              end
             else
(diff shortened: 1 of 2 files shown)


jkeen:
@zailleh Is this PR still relevant? I'm trying to get up to speed here, haven't run into this issue personally, and am not fully clear if #354 (merged in 2021) took care of this.

zailleh:
> @zailleh Is this PR still relevant? I'm trying to get up to speed here, haven't run into this issue personally, and am not fully clear if #354 (merged in 2021) took care of this. 

Yeah it looks like the same issue so we can close this one!

PR #310, opened 2021-01-30
Add validation error on missing sidepost

richmolj:
[This issue](https://github.com/graphiti-api/graphiti-rails/issues/65) raised the idea that we should be treating a missing sideposted record as a validation, not a 401.

In the course of this work I found [this older PR](https://github.com/graphiti-api/graphiti/pull/214), which initially raised the same point in [this comment](https://github.com/graphiti-api/graphiti/pull/214#issuecomment-597756121).

I very much agree that this should be a validation error, not a 401, but it breaks backwards compatibility. So, you can opt-in to this behavior with `Graphiti.config.raise_on_missing_sidepost = false`. We should consider this a best practice going forward and note it in the blog.

### DIFF
--- a/lib/graphiti/adapters/persistence/associations.rb
+++ b/lib/graphiti/adapters/persistence/associations.rb
@@ -11,8 +11,16 @@ module Graphiti
                   .persist_with_relationships(x[:meta], x[:attributes], x[:relationships])
                 processed << x
               rescue Graphiti::Errors::RecordNotFound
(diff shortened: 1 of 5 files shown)


richmolj:
@ef4 does this work for you?

/cc @jackcasey @aribouius who were involved in the prior issue https://github.com/graphiti-api/graphiti/pull/214#issuecomment-597756121

wadetandy (approved):

ef4:
Looks good, thanks.

PR #313, opened 2021-02-19
Sort remote resources in schema generation

PChambino:
Remote resources are not ordered which causes the schema to randomly change depending on load order.


richmolj:
Ah this is great thanks @PChambino - would you mind adding a tiny spec to ensure we don't regress?

PChambino:
Added the spec. I think that CI failure is unrelated and transient. 🤔

richmolj:
Yup think we're good here. Thanks @PChambino !

PR #314, opened 2021-02-21
Support polymorphic includes

richmolj:
Currently for polymorphic resources we just check if the include is supported - if it is, render it. For instance in the sample app
`Epic` is a type of `Task` and it's the only one with the `milestones` relationship. So `/tasks?include=milestones` works like
you'd think.

The issue is if both `Feature` and `Epic` support the `milestones` relationship (or if all types do via the parent resource), you can't conditionally request milestones just for epics. Enter this PR.

We introduce a special `on__<type>--<name>` syntax to conditionally include based on type, e.g. `?include=on__epics--milestones`. We can now support the use case. Note the syntax here is ugly, but we're tied to a finite set of supported characters according to the JSON:API spec.

In practice I view this as a kind of internal API for supporting GQL requests, so I'm only testing the hash renderer. I don't really want to advertise it unless we get requests to support it for jsonapi.

Note we're only doing this for includes, not fields. For fields we can reference the jsonapi type as normal and this would only be an issue if double-listing the same entities as multiple levels of the graph. Maybe we support this one day, but for now I'd like to limit the complexity/feature to just this case.

Finally - yes, the hash renderer is getting pretty ugly. Hands a little tied by the way jsonapi-rb works. I think we're nearing max ugliness and may also improve performance by writing our own (still needs to use serialization logic from jsonapi-rb), but holding off for now.

### DIFF
--- a/lib/graphiti/query.rb
+++ b/lib/graphiti/query.rb
@@ -96,7 +96,18 @@ module Graphiti
               sl_resource = resource_for_sideload(sideload)
               query_parents = parents + [self]
               sub_hash = sub_hash[:include] if sub_hash.key?(:include)
-              hash[key] = Query.new(sl_resource, @params, key, sub_hash, query_parents, :all)
+
+              # NB: To handle on__<type>--<name>
+              # A) relationship_name == :positions
+              # B) key == on__employees.positions
+              # This way A) ensures sideloads are resolved
+              # And B) ensures nested filters, sorts etc still work
+              relationship_name = sideload ? sideload.name : key
+              hash[relationship_name] = Query.new sl_resource,
+                @params,
+                key,
+                sub_hash,
+                query_parents, :all
             else
               handle_missing_sideload(key)
             end
--- a/lib/graphiti/resource/polymorphism.rb
+++ b/lib/graphiti/resource/polymorphism.rb
@@ -42,9 +42,14 @@ module Graphiti
         end
 
         def sideload(name)
-          sl = super
+          if (split_on = name.to_s.split(/^on__/)).length > 1
+            on_type, name = split_on[1].split("--").map(&:to_sym)
+          end
+
+          sl = super(name)
           if !polymorphic_child? && sl.nil?
             children.each do |c|
+              next if on_type && c.type != on_type
               break if (sl = c.sideloads[name])
             end
           end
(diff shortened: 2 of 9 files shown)


richmolj:
@wadetandy

lib/graphiti/query.rb
```diff
@@ -90,6 +90,9 @@ def sideloads
       @sideloads ||= begin
         {}.tap do |hash|
           include_hash.each_pair do |key, sub_hash|
+            if key.to_s.starts_with?('on__')
```
wadetandy: We are parsing this key out in both the query and the serializer. Maybe extract a small `Graphiti::Util::PolymorphicIncludeDirective` class which does the string splitting and gives you the parts?

wadetandy (approved):
One small cleanup comment. If I'm reading right, one drawback of this implementation is if I have two polymorphic children with the same relationship but I only ask to include one, both will be resolved and only one will be included in the payload? That's the status quo though, so we're at least saving serialization costs.

lib/graphiti/query.rb
richmolj: I think this is a good comment, but I'm going to hold off on it for now to avoid the indirection. Maybe if we touch it again or add fields support.

richmolj:
> f I'm reading right, one drawback of this implementation is if I have two polymorphic children with the same relationship but I only ask to include one, both will be resolved and only one will be included in the payload? 

Made a tiny refactor to make this more clear (logic moves from `query.rb` to `resource/polymorphism.rb`). Only the specific subtype will be resolved AND rendered. Worth noting the scenarios in the epic/feature/milestone case - we'd file 2 queries, one filtering by `epic_id` and another by `feature_id`.

lib/graphiti/resource/polymorphism.rb
```diff
@@ -42,9 +42,14 @@ def inherited(klass)
         end
 
         def sideload(name)
-          sl = super
+          if (split_on = name.to_s.split("on__")).length > 1
```
wadetandy: Probably overkill, but how about splitting on a regex to limit to string start: /\Aon__/

richmolj: Done!

richmolj:
Added a few more specs and addressed feedback - might touch more if we decide to accomodate fields, but think we're good for now

PR #316, opened 2021-02-22
Raise error when unreadable attribute in GQL context

richmolj:
Our current behavior is to simply not render the attribute when the
readable guard fails, but to adhere to the GQL spec we will instead
need to raise. There's a whole rat's nest of different auth patterns,
but this is the most straightforward thing for now.


wadetandy (approved):
Yeah seems like the most straightforward approach and unblocks until we can consider more advanced versions.

PR #319, opened 2021-02-26
Allow same name for attribute and relationship

richmolj:
This was already possible unless the attribute was guarded, in which case the relationship would incorrectly share the same guard. This is a "bug" in jsonapi-rb, so we add to our existing hacks on top of that library pending an upstream fix.

In practice this is helpful for a GQL-specific federated attribute, which would share the same name as a relationship but be conditionally rendered only for GQL.


wadetandy (approved):
Nice tests.

PR #327, opened 2021-03-09
Allow access to extra_fields with polymorphic parent resource type

revdan:
Following up from [this draft](https://github.com/graphiti-api/graphiti/pull/326) yesterday.

We're using polymorphism, imagine the following class with an extra field on the top level:
```
class ThingResource < ApplicationResource
  self.polymorphic = [
    'Things::RedThing',
    'Things::BlueThing'
  ]

  extra_attribute :type, :string
```

If I were to visit `example.com/api/things/1`, it could be a `RedThing` or `BlueThing` - I don't know which. 

But in the current implementation, `example.com/api/things/1?extra_fields[things]=type` wouldn't work, as the `@_type` var is the type of the child. I'd have to specify either `?extra_fields[red_thing]=type` or `?extra_fields[blue_thing]=type` to get past the guard clause.

### DIFF
--- a/lib/graphiti/extensions/extra_attribute.rb
+++ b/lib/graphiti/extensions/extra_attribute.rb
@@ -46,9 +46,9 @@ module Graphiti
               next false unless instance_exec(&options[:if])
             end
 
-            @extra_fields &&
-              @extra_fields[@_type] &&
-              @extra_fields[@_type].include?(name)
+            next false unless @extra_fields
+
+            @extra_fields[@_type]&.include?(name) || @extra_fields[@resource&.type]&.include?(name)
           }
 
           attribute name, if: allow_field, &blk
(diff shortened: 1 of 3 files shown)


richmolj:
Looks great @revdan, thanks so much ❤️   Would you mind making those `before do...end` blocks instead of `before {}` blocks, just to match the other specs? And ideally rebase into one commit? After that I think we're GTG! 🚀

revdan:
> Looks great @revdan, thanks so much ❤️ Would you mind making those `before do...end` blocks instead of `before {}` blocks, just to match the other specs? And ideally rebase into one commit? After that I think we're GTG! 🚀

Brilliant, will do. Thanks @richmolj 😄

revdan:
Should be good now @richmolj 🤞

richmolj:
Perfect, thank you @revdan 💯 🙏 !

PR #329, opened 2021-03-09
Propagate nested extra_fields to links

### LINKED ISSUE #328: Propagating extra_fields to related links
Current Graphiti won't propagate the `extra_fields` query parameter for nested resources in their related links.

For example, for a request to `http://example.com/books?extra_fields[authors]=biography` the response would contain in `data.relationship.author.links.related` the following URL: `http://example.com/authors/<author_id>` and does not propagate the `extra_fields` parameter to the nested/child resource (i.e. the URL is not `http://example.com/authors/<author_id>?extra_fields[authors]=biography`).

This is a feature which we are currently implementing by setting the `params_proc` on relationships along with the following monkey-patch on `Resource#around_scoping`:
```ruby
Graphiti::Resource.prepend(Module do
  def around_scoping(scope, query_hash)
    if (extra_fields = query_hash.dig(:extra_fields, type))
      query_hash[:extra_fields][type] = extra_fields.split(',') unless extra_fields.is_a?(Array)
    end

    super(scope, query_hash)
  end
end)
```

The monkey-patch is required in order to avoid serialising `extra_fields` as array query parameters in `Util::Link#raw_url` (i.e. encoding as `extra_fields[resource][]=a&extra_fields[resource][]=b` rather than `extra_fields[resource]=a,b`.

Is there scope for making propagation of `extra_fields` a part of the default propagated parameters for links?
Potential within `Util::Link#params` in the same way the `:filter` is propagated?

CGA1123:
Propagates the `extra_fields` attributes for a nested resource to the
relationship links created.

e.g. `GET http://example.com/books/1?extra_fields[authors]=biography`
should return a link to the `author` looking like
`http://example.com/authors/2?extra_fields[authors]=biography`,
propagating the `extra_fields` that apply for the nexted `authors`
resource.

Closes #328

### DIFF
--- a/lib/graphiti/sideload.rb
+++ b/lib/graphiti/sideload.rb
@@ -136,6 +136,14 @@ module Graphiti
       base_filter(parents)
     end
 
+    def link_extra_fields
+      extra_fields_name = [association_name, resource.type].find { |param|
+        context.params.dig(:extra_fields, param)
+      }
+
+      {resource.type => context.params.dig(:extra_fields, extra_fields_name)} if extra_fields_name
+    end
+
     # The parent resource is a remote,
     # AND the sideload is a remote to the same endpoint
     def shared_remote?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 Features:
+- [329](https://github.com/graphiti-api/graphiti/pull/329) Propagate `extra_fields` to related resource links.
 - [242](https://github.com/graphiti-api/graphiti/pull/242) Bump `jsonapi-renderer` to `~0.2.2` now that (https://github.com/jsonapi-rb/jsonapi-renderer/pull/36) is fixed.
 - [158](https://github.com/graphiti-api/graphiti/pull/158) Filters options `allow_nil: true`
   Option can be set at the resource level `Resource.filters_accept_nil_by_default = true`. 
--- a/lib/graphiti/util/link.rb
+++ b/lib/graphiti/util/link.rb
@@ -63,6 +63,10 @@ module Graphiti
             params[:filter] = @sideload.link_filter([@model])
           end
 
+          if (extra_fields = @sideload.link_extra_fields)
+            params[:extra_fields] ||= extra_fields
+          end
+
           @sideload.params_proc&.call(params, [@model], context)
         end
       end
--- a/spec/serialization_spec.rb
+++ b/spec/serialization_spec.rb
@@ -1322,6 +1322,66 @@ RSpec.describe "serialization" do
           end
         end
 
+        context "with nested extra attribute" do
+          context "when referencing the type" do
+            let!(:employee) { PORO::Employee.create(classification_id: 789) }
+
+            before do
+              resource.belongs_to :classification
+              params[:extra_fields] = {classifications: "description"}
+            end
+
+            def classification
+              json["data"][0]["relationships"]["classification"]
+            end
+
+            it "links correctly" do
+              render
+              expect(classification["links"]["related"])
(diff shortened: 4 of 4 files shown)


lib/graphiti/sideload.rb
```diff
@@ -136,6 +136,12 @@ def link_filter(parents)
       base_filter(parents)
     end
 
+    def link_extra_fields
+      extra_fields = context.params.dig(:extra_fields, name)
```
CGA1123: I'm not sure whether `name` is the correct attribute to use in general, it seems correct in the specs.

richmolj: This is going to be the tricky part. `name` is `Sideload#name`, or IOW the relationship name as opposed to the JSON:API type. So imagine `PositionResource` had `self.type = :special_positions`, the relationship name is `positions` but the JSON:API type is `special_positions`.

Adding a `belongs_to` spec would catch this, since the relationship name (`department`) wouldn't match the type (`departments`).

The complicating factor here is there are 2 parts - getting the fields from the url (line 140) and building the link (142). I think it makes sense that the latter is always the JSON:API type (I think here it would be `self.resource.type`). This is why the `belongs_to` spec would fail - you'd end up with `?extra_fields[department]` instead of `?extra_fields[departments]`.

But to get the fields from the URL, there are actually two few different ways you could specify this - one is the type, one is the name. The former is typical JSON:API spec, the latter is useful for fine-grained nesting (ie we want to target `positions.department.tasks.employee` not the top-level Employee or any other employee in the graph). So we'd need specs ensuring that this works however you specify it. See the specs on `query.rb` for more context if needed.

CGA1123: Thanks for the pointers, I ended up added `name`, `association_name`, or `resource.type` as the potential ways to reference the related resource. Does that that make sense? Or should `association_name` and `resource.type` alone be enough?

Are there any scenarios where the `association_name` (`as:`) is set and someone would refer using the `name` directly? 🤔

richmolj: Really good question, thanks for bringing it up. I think `association_name` and `resource.type` should be enough. The use case is polymorphism, so `has_many :notes, as: :notable`. In the actual specs I believe this is `has_many :employees, as: :workspace`. I believe this will "just work" but would you mind adding a spec for this as well?

CGA1123: Okay, thanks for clarifying!

There is already a test case that I think covers the `has_many :employees, as: :workspace` case:
https://github.com/graphiti-api/graphiti/blob/c866b230fe69ce08dca2eaef5116c3f32b859009/spec/serialization_spec.rb#L1345-L1356

Does that cover the same case you are referring to?

richmolj: Yep, exactly. We just want to make sure the correct link is generated for this extra fields use case. Then gtg I think!

CGA1123: Ah, I understand now! The `polymorphic_belongs_to` case is what needs to be verified.
Whereby if the request has `?extra_fields[workspace]=foo` we should create a link with `?extra_fields[office]=foo` or `?extra_fields[home_office]=foo` depending on which underlying resource backs the association. 👍

richmolj: Yep I think you got it 👏 !

CGA1123: Sorry for the delay 😄 

I've added the test case using the `credit_card` example as the workplace example is set up is on the `ActiveRecord` resources and not within the `PORO` ones.

richmolj:
You're awesome @CGA1123, thanks for sticking with it 💯 ! Thanks again for the PR ❤️ !

CGA1123:
@richmolj Thanks for the help! 🚀

PR #334, opened 2021-03-11
Added any_required option to filter attributes

chiliburger:
Added `any_required` option to filter attributes when we want one out of a set of attribute filters required. For example if we want to filter by either `employee_id` or `employee_uid`  field. This commit raises an error if one of the `any_required` attribute filters are not passed into the request.

Question: Should we only apply the first of any_required attributes? or go ahead and apply all as long as one is present. Currently it will apply all as long as one attribute `any_required` is present.

### DIFF
--- a/lib/graphiti/errors.rb
+++ b/lib/graphiti/errors.rb
@@ -690,7 +690,7 @@ module Graphiti
       end
 
       def message
-        "#{@resource.class.name}: One of the following filters must be passed in: #{@filters.keys.join(', ')}"
+        "#{@resource.class.name}: One of the following filters must be passed in: #{@filters.keys.join(", ")}"
       end
     end
 
--- a/lib/graphiti/errors.rb
+++ b/lib/graphiti/errors.rb
@@ -683,6 +683,17 @@ module Graphiti
       end
     end
 
+    class MissingAnyRequiredFilter < Base
+      def initialize(resource, filters)
+        @resource = resource
+        @filters = filters
+      end
+
+      def message
+        "#{@resource.class.name}: One of the following filters must be passed in: #{@filters.keys.join(', ')}"
+      end
+    end
+
     class ResourceNotFound < Base
       def initialize(resource_class, sideload_name, tried)
         @resource_class = resource_class
--- a/spec/filtering_spec.rb
+++ b/spec/filtering_spec.rb
@@ -1783,4 +1783,56 @@ RSpec.describe "filtering" do
       expect(records.map(&:id)).to eq([employee2.id])
     end
   end
+
+  context "when filter is any_required on .attribute" do
+    before do
+      resource.attribute :first_name, :string, filterable: :any_required
+      resource.attribute :last_name, :string, filterable: :any_required
+    end
+
+    context "when one is given in the request" do
+      before do
+        params[:filter] = {first_name: "Agatha"}
+      end
+
+      it "works" do
+        expect(records.map(&:id)).to eq([employee2.id])
+      end
+    end
+
+    context "when none are given in request" do
+      it "raises error" do
+        expect {
+          records
+        }.to raise_error(/One of the following filters must be passed in: first_name, last_name/)
+      end
+    end
+  end
+
+  context "when filter is any_required on .filter", focus: true do
+    before do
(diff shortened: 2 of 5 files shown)


lib/graphiti/errors.rb
```diff
@@ -683,6 +683,17 @@ def message
...
     end
 
+    class MissingAnyRequiredFilter < Base
+      def initialize(resource, filters)
+        @resource = resource
+        @filters = filters
+      end
+
+      def message
+        "#{@resource.class.name}: One of the following filters must be passed in: #{@filters.keys.join(', ')}"
```
richmolj: Standard would prefer double quotes here `.join(", ")`

spec/filtering_spec.rb
```diff
@@ -1783,4 +1783,56 @@ def after_filtering(scope)
...
+    context "when none are given in request" do
+      it "raises error" do
+        expect {
+          records
+        }.to raise_error(/One of the following filters must be passed in: first_name, last_name/)
+      end
+    end
+  end
+
+  context "when filter is any_required on .filter", focus: true do
```
richmolj: Looks like a leftover `focus`

spec/filtering_spec.rb
```diff
@@ -1783,4 +1783,56 @@ def after_filtering(scope)
...
+    before do
+      resource.config[:filters] = {}
+      resource.config[:attributes] = {}
```
richmolj, line 1815: 👍

richmolj (commented):
Wonderful PR, thanks so much @chiliburger ❤️ ! Valid use case that I hadn't thought of, implemented well.

I had some initial thoughts wrt the `:dependent` option, which is *kind of* the corresponding `all` to this PR's `any`. But ultimately I like the simplicity of following the same existing `required` pattern.

I think we should probably update the schema as well, you can see [these specs for vanilla required](https://github.com/graphiti-api/graphiti/blob/master/spec/schema_spec.rb#L379-L417).

There are couple specs failing in CI that aren't your fault since you don't see locally without knowing to run them. One is a failure by the `standard` linter, the other is that `Array#intersection` only exists in ruby 2.7+.

> Should we only apply the first of any_required attributes?

If I'm understanding correctly - both filters are passed in the request, should we apply only one or both? I think you're correct that it should be both. There might be separate work with a `deprecated` flag (which I think would be a common use case here) down the line, where we could choose to drop the deprecated one.

Again thanks so much, this is awesome 👏

chiliburger:
Thanks for taking a look at this @richmolj. I will work on cleaning up the TravisCI issues and remove the focus options (ooops).

wadetandy (commented):
Hey @chiliburger thanks for this. I think the functionality is really nice here, but I don't personally love that this lives on the attribute DSL.  The fact that it links one attribute to another in an implicit way (especially if you're dealing with inheritance where there might be an `any_required` across parent and child classes).

If there was a small additional DSL dedicated to this concept I think it might read a lot more clearly and set us up for future enhancement to this type of behavior:

```ruby
class MyResource < ApplicationResource
    attribute :first_name, :string, filterable: true
    attribute :last_name, :string, filterable: true
    
    filter_group [:first_name, :last_name], required: :any 
end
```

This allows these to be tied together more explicitly while also opening up space for things like only allowing filtering on one or the other, etc.

chiliburger:
Thanks for the feedback @wadetandy! I am pretty new to the Graphiti framwork but that does make sense and I like the syntax. I will look at this option some more.

richmolj:
I think I agree with what @wadetandy is saying (inheritance is a good point), and it's part of where I was going mentioning 'dependent' above.

But I still think the PR works as-is as an incremental approach. The attributes are not necessarily coupled; you could have a single 'any_required' that works just like OG 'required'. One doesn't need to know about the other. I also like that this is just following how 'required' already works - I can see how it would make a lot of sense to move out of the attribute DSL (so it doesn't conflict with guards) but that horse has already left the barn.

IOW I support both approaches, and I'll even go ahead and favor Wade's a bit, but either works for me.

chiliburger:
@richmolj @wadetandy closing in favor of PR https://github.com/graphiti-api/graphiti/pull/336.

PR #335, opened 2021-03-11
Upgrade to Standard 1.0

wadetandy:
Now that this has gone 1.0, we shouldn't see any breaking changes going
forward, so let's get up to that baseline.


wagenet:
I rebased this and ran the auto-fix again.

richmolj:
Holy crap thank you

PR #336, opened 2021-03-12
Added filter_group DSL:

chiliburger:
Added filter_group DSL to require filtering by all or any of the
filters specified in the filter group. For example if we want to
filter by either employee_id or employee_uid field.

This commit raises an error if any/all of the required filters
are not passed into the request.

filter_group: [:filter1, filter1], required: :all | :any


chiliburger:
Manually tested functionality with existing service. Need to add unit tests and add handling of new error classes in Graphiti Rails.

lib/graphiti/scoping/filter_group.rb
```diff
@@ -0,0 +1,81 @@
+module Graphiti
+  # @api private
+  module Scoping::FilterGroup
```
wadetandy: I know you're still in draft mode and the functionality looks awesome so far. Wanted to jump in an offer one comment on this though before you get too far with it. 

Seems from quickly reading through that the only dependency on the `Scoping::Filter` class that you are mixing into is the `resource` and `query_hash` methods. What do you think about instead making this a `Scoping::FilterGroupValidator` class which accepts the resource as a collaborator? This could still be called in the same location with `FilterGroupValidator.new(resource, query_hash).raise_unless...`, but would probably be a bit easier to test behavior and have less indirection in the form of a mixin as well.

lib/graphiti/errors.rb
```diff
@@ -816,5 +816,34 @@ def message
...
+          #{@valid_required_values.join(", ")}
+        MSG
+      end
+    end
+
+    class FilterGroupMissingRequiredFilters < Base
+      def initialize(resource, filter_names, required)
+        @resource = resource
+        @filter_names = filter_names
+        @required_label = required == :all ? 'All' : 'One'
```
gfmurphy: Do we want to extend this concept to describe and "exclusive or" state? i.e. we want to accept one or the other filter, but not both. I could imagine we could need that as well:

1. `all` -> and
2. `any` -> or
3. `?` -> xor

Does that make sense to you @chiliburger?

wadetandy: I think it’s a low hanging fruit for enhancement if and when we need it, but don’t think we need to hold up this feature getting merged either. I'll leave that up to @chiliburger to decide whether to include or not. As for name, if we do add it, I'd suggest `required: :one` or `required: :only_one`

lib/graphiti/scoping/filter_group.rb
chiliburger: @wadetandy that sounds good. This is the sort of feedback I was hoping for posting the draft.

lib/graphiti/errors.rb
chiliburger: Yes I considered a `:detect` option like in Ruby Enumerable but it would make the implementation a bit more complex to skip filters. Trying to keep it simple at the moment.

lib/graphiti/scoping/filter_group_validator.rb
```diff
@@ -0,0 +1,82 @@
+module Graphiti
+  class Scoping::FilterGroupValidator
+    VALID_FILTER_GROUP_REQUIRED_VALUES = %i[all any]
+
+    def initialize(resource, query_hash)
+      @resource = resource
+      @query_hash = query_hash
+    end
+
+    def raise_unless_filter_group_requirements_met?
```
wadetandy: Can we make these `bang!` methods instead of `predicate?` methods?

chiliburger:
@wadetandy are there other specs/docs that need to be modified beyond the spec/filtering_spec.rb file?

spec/filtering_spec.rb
```diff
@@ -1783,4 +1783,73 @@ def after_filtering(scope)
...
+
+      context "when none are given in the request" do
+        it "raises an error" do
+          expect {
+            records
+          }.to raise_error(/One of the following filters must be provided on resource/)
+        end
+      end
+
+      context "when one is given in the request" do
```
wadetandy, line 1840: Would add a test to ensure it doesn't blow up with both provided.

wadetandy:
No I think these tests are good from an integration perspective. Might want to add at least a couple unit tests for the new validator class. Should just be able to use structs or doubles for the resource/query_hash.

richmolj:
This is *awesome* @chiliburger, thanks for going the extra mile 💯 🥇 👏 ! I did just want to raise the issue that we might want to think about how this should appear in the schema/backwards-compatibility specs. But this looks great 👍

chiliburger:
> No I think these tests are good from an integration perspective. Might want to add at least a couple unit tests for the new validator class. Should just be able to use structs or doubles for the resource/query_hash.

@wadetandy Added the suggested specs.

lib/graphiti/errors.rb
```diff
@@ -816,5 +816,34 @@ def message
...
+
+    class FilterGroupInvalidRequirement < Base
+      def initialize(resource, valid_required_values)
+        @resource = resource
+        @valid_required_values = valid_required_values
+      end
+
+      def message
+        <<-MSG.gsub(/\s+/, " ").strip
+          The filter group required: value on resource #{@resource.class} must be one of the following:
```
wadetandy, line 828: how about `@resource.jsonapi_type` so that it more closely matches the url param? This will provide users with actionable error messages once we wire these up to graphiti-rails error handling.

(This might be wrong if my comment on the validator class below is correct, as this is more a message for API devs, not users. Would still apply to the other error message below though)

lib/graphiti/errors.rb
```diff
@@ -816,5 +816,34 @@ def message
...
+    class FilterGroupMissingRequiredFilters < Base
+      def initialize(resource, filter_names, required)
+        @resource = resource
+        @filter_names = filter_names
+        @required_label = required == :all ? "All" : "One"
+      end
+
+      def message
+        <<-MSG.gsub(/\s+/, " ").strip
+          #{@required_label} of the following filters must be provided on resource #{@resource.class}:
```
wadetandy: `jsonapi_type`

lib/graphiti/scoping/filter_group_validator.rb
```diff
@@ -0,0 +1,76 @@
...
+    def raise_unless_filter_group_requirements_met!
+      return if grouped_filters.empty?
+
+      case filter_group_requirement
+      when :all
+        raise_unless_all_requirements_met!
+      when :any
+        raise_unless_any_requirements_met!
+      else
+        raise Errors::FilterGroupInvalidRequirement.new(
```
wadetandy: If I'm reading this correctly, this is raised if I declare the filter group incorrectly?

```ruby
filter_group [:foo, :bar], no_required_key: :all
```

If so, I think we should do that as part of the DSL method at initialization time instead of at runtime.

wadetandy (changes requested):
A couple small comments but this is looking fantastic and is a great addition. Excited to fix these small things and get this merged!

lib/graphiti/scoping/filter_group_validator.rb
chiliburger: Yes correct, raises the error if the :required option is invalid. Raising at DSL method initialization sounds good.

lib/graphiti/errors.rb
chiliburger: Yes this error is more for devs in the case of passing in incorrect values during dev.

lib/graphiti/errors.rb
chiliburger: Sounds good, this error is client facing. Thanks for the tip!

chiliburger: @wadetandy changed to use the `type` method on the resource.

lib/graphiti/scoping/filter_group_validator.rb
chiliburger: @wadetandy moved this error check to the DSL method invocation.

wadetandy (approved):
This is great stuff! Thanks for hanging in there with all the comments back and forth. Really pleased with how this turned out.

wadetandy:
https://rubygems.org/gems/graphiti/versions/1.2.34

richmolj:
This is awesome and I have to echo @wadetandy's thoughts - thanks for sticking with it to end up at a great place @chiliburger  🙏 

Created an issue for adding this concept to the schema/diffing, which is work we can follow up with later https://github.com/graphiti-api/graphiti/issues/338

chiliburger:
Thanks @wadetandy and @richmolj! I appreciate the collaboration, constructive feedback and fast response times with this PR.

Issue #339, opened 2021-03-16
Missing "type" field fails confusingly


ef4:
If a client forgets to include the required `type` field in a JSON:API POST, instead of giving a validation error graphiti treats the payload as empty. 

It looks like the body serializes to empty here:

https://github.com/graphiti-api/graphiti/blob/054685c29d8641388a16787591afec2c62f62244/lib/graphiti/request_validators/validator.rb#L38-L42

It would be much better to error loudly that the request is not valid because it lacks a type. It took me quite a long time to figure out what was happening. 😝

richmolj:
Oof, sorry for the hassle, I quite agree. The linked code is even in a RequestValidator class! Thanks for the heads up @ef4

richmolj:
Fix has been merged, thanks again @ef4 for raising the issue!

PR #340, opened 2021-03-18
Add validation error when data/type missing

richmolj:
Addresses https://github.com/graphiti-api/graphiti/issues/339

A few other things came up - one is that all requests, even reads, go
through the validator. Ideally the check is further upstream but I
tried to change the least amount of code while making this explicit.

The other is that I'm normalizing the params in the constructor, so
we can always deal with something normalized instead of worrying about
raw vs normalized. This has a downstream effect of changing the error
message for a Rails4-specific test you can see in the diff (if
anything, probably more correct now).


wadetandy:
Nice ✔️

wadetandy (approved):

Issue #341, opened 2021-03-18
Bypass filter_group on find requests


chiliburger:
Required filters are bypassed on find requests but the filter_group is still applied returning a 400 on find requests if the filters are not passed in. Need to bypass the filter_group requirements as well on find requests.

PR #342, opened 2021-03-18
filter_group bypass fix

chiliburger:
Added filter_group bypass for resource find method.

When adding the filter_group dsl it did not respect the bypass_required_filters
 flag. This PR changes the filter_group to respect the flag.

Issue: https://github.com/graphiti-api/graphiti/issues/341


richmolj (approved):
CI taking its sweet time but this looks good when it's done. Sorry we didn't catch this in review and thanks for the fix ❤️

richmolj:
Released in `1.2.35`

chiliburger:
> CI taking its sweet time but this looks good when it's done. Sorry we didn't catch this in review and thanks for the fix ❤️

@richmolj No worries! I should have been testing with a find action. Thanks for jumping on this!

Issue #338, opened 2021-03-16
Add filter_group to schema/diffing


richmolj:
We should add the new `filter_group` to the schema so clients know what filters are required. We already have vanilla `required` in the schema, and [we'll raise a backwards-compatibility error](https://github.com/graphiti-api/graphiti/blob/master/spec/schema_diff_spec.rb#L709-L729) if a filter goes from required to not. In addition, this will allow us to implement the same functionality in in our alpha GraphQL project, which builds a GQL schema from a Graphiti schema.

This probably involves adding a new node to the schema at the same level as `filters.`

PR #343, opened 2021-03-19
Add filter_group to schema/diffing

richmolj:
Scenarios for diffing are best outlined in the spec, but should be anything that would cause a client to break (for instance adding to the group when all are required).

### DIFF
--- a/lib/graphiti/schema.rb
+++ b/lib/graphiti/schema.rb
@@ -99,6 +99,10 @@ module Graphiti
           relationships: relationships(r)
         }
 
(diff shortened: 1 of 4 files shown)


richmolj:
Addresses https://github.com/graphiti-api/graphiti/issues/338

wadetandy (approved):

richmolj:
@chiliburger just a heads up

chiliburger:
Nice work! Thanks for handling this!

PR #348, opened 2021-03-24
Use blank? instead of nil? for typecasting

cubasepp:
If I want to update an `attribute :birth_date, :datetime` with `""` I run into a typecasting error.
For me such validation should be done later, if the filed is required...

In my opinion this JSON should be treated equally
```
{
  "value_1": "",
  "value_2": null
}
```

So start typecasting only if the value is not blank.

### DIFF
--- a/lib/graphiti/resource.rb
+++ b/lib/graphiti/resource.rb
@@ -73,7 +73,7 @@ module Graphiti
         type_name = filters[name][:type]
       end
       type = Graphiti::Types[type_name]
-      return if value.nil? && type[:kind] != "array"
+      return if value.blank? && type[:kind] != "array"
       begin
         flag = :read if flag == :readable
         flag = :write if flag == :writable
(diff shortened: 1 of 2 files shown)


richmolj:
@cubasepp thanks for the PR ❤️ ! But sadly I'm not sure I agree. Or maybe better said - I think I'm on the fence and worried about backwards-compatibility for something I'm on the fence about.

At minimum we'd need specs ensuring that `""` doesn't end up in the DB. Even past that, we have the backwards-compatibility issue. Is scrubbing these client-side to `null` not an option for you?

cubasepp:
@richmolj  of course it is possible to handle it on the frontend as well. You have always manually  transform the `<input>` values to `null` for `attributes` which needs a typecast.

This was the easiest solution for me. 

Regarding backwards-compatibility, what do you think about an additional config params like `no_typecast_on_blank: true`?

And put this into `lib/graphiti/request_validators/validator.rb`

```
def typecast_attributes
  ...
  next if value.blank? && resource.class.config[:attributes][:key][:no_typecast_on_blank]
end
```

cubasepp:
Okay I see the problem with backwards-compatibility.

I will make changes. This can be ignored, also the tests fail of this problem.

PR #349, opened 2021-03-25
Make typecast on blank values configurable

cubasepp:
Example: `attribute :birthdate, :datetime, typecast_on_blank: false`
This option allows you to pass either `null` or `""` as value to set empty values without typecast.

The JSON should be treated equally:
```
{
  "value": "",
  "value": null
}
```

### DIFF
--- a/lib/graphiti/request_validators/validator.rb
+++ b/lib/graphiti/request_validators/validator.rb
@@ -77,6 +77,10 @@ module Graphiti
             key == :id &&
             resource.class.config[:attributes][:id][:writable] == false
 
+          if resource.class.config[:attributes].has_key?(key)
+            next unless resource.class.config[:attributes][key][:typecast_on_blank] || value.present?
+          end
+
           begin
             attributes[key] = resource.typecast(key, value, :writable)
           rescue Graphiti::Errors::UnknownAttribute
--- a/lib/graphiti/resource/configuration.rb
+++ b/lib/graphiti/resource/configuration.rb
@@ -88,6 +88,7 @@ module Graphiti
           :attributes_sortable_by_default,
           :attributes_filterable_by_default,
           :attributes_schema_by_default,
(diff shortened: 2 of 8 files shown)


cubasepp:
@richmolj local the test are running. Maybe we can retrigger it.  I hope you like this approach more :)

richmolj:
I think the PR is great for what you are trying to do. I'm not sure I am a fan of adding - and supporting - yet another option, though. I also remain skeptical of the functionality. For a string field, there could a legitimate different between `""` (there is no middle name) and `null` (we don't know if there is a middle name).

If a number of users were asking for this I could be convinced otherwise.

For now, I wonder if this solves your issue in a clean way:

```ruby
class ApplicationResource < Graphiti::Resource
  # ... code ...

  def typecast(key, value, flag)
    value.blank? ? nil : super
  end
end
```

cubasepp:
No problem if you don't like it. :)

Overriding the `typecast` solves my problem so I have an easy clean way - thanks. 🎉 
But only for all `attributes` not only fo a specified, but I can live with this limitation atm. :)

>  I also remain skeptical of the functionality. For a string field, there could a legitimate different between "" (there is no middle name) and null (we don't know if there is a middle name).

What I can do is, not apply this option for type string. 
At least you will pass always the typecast for type string. You will never get an error: `should be type string`  

Check test or this here :)
```
attribute :value1, :string
attribute :value2, :string
attribute :value3, :string

{
   "value1": 0,
   "value2": true,
   "value3": "String"
}
```

So if I cannot convince you I will close this PR :)

Thanks for the fast response and happy weekend.

richmolj:
Thanks for your flexibility @cubasepp ❤️  I think your ideas are good, but it probably makes sense for us to limit the features we need to support unless more users start clamoring for it. Glad you have a workaround for now! Will close this but keep it in mind if it comes up again.

PR #350, opened 2021-03-26
Allow unwritable PK when associating on create

richmolj:
Imagine updating a `Position`, changing the `Employee` it is associated
with. Since we're justassociating, we don't need `Employee`s PK to be
writable. But because our internals end up saving the `Employee`
regardless - we just walk the tree, and `id` is part of the attributes
considered - so we'd normally validate the PK, even though we really
don't need to.

We already avoided validating the PK when *updating* the `Position`,
this PR also considers *creating* the `Position` and associating to a
pre-existing `Employee`. We do this by looking at the `method` of
each sidepost, instad of the overall action being called.

### DIFF
--- a/spec/integration/rails/persistence_spec.rb
+++ b/spec/integration/rails/persistence_spec.rb
@@ -404,8 +404,6 @@ if ENV["APPRAISAL_INITIALIZED"]
     end
 
     describe "non-writable foreign keys" do
-      subject(:make_request) { do_update(payload) }
-
       context "when belongs_to" do
         let(:klass) do
           Class.new(EmployeeResource) do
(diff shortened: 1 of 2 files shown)


spec/integration/rails/persistence_spec.rb
```diff
@@ -491,10 +520,29 @@ def self.model_name
...
+        end
+
+        context "and the overall request is update" do
+          subject(:make_request) { do_update(payload) }
+
+          before do
+            payload[:data][:id] = employee.id.to_s
+          end
+
+          it "does not require the FK to be a writable attribute" do
```
wadetandy, line 542: The title of the PR says "when associating on create" but this allows it on update too. Just checking that was intended.

wadetandy: Nevermind, re-read the PR description and it makes sense.

wadetandy (approved):

PR #352, opened 2021-04-13
Don't blow up when missing/unexpected context

richmolj:


richmolj:
@wadetandy just heads up

PR #354, opened 2021-04-23
Fix many_to_many validation exception when sideposting

KirchoKirev:
This fix is related to previously opened PR - https://github.com/graphiti-api/graphiti/pull/304.

When adding many_to_many related associations - for example `department.teams` << `new_team`, `save!` gets called on the related `new_team` and it throws a `ValidationError` that breaks the normal graphiti error handling. Prevent this by checking if the association is valid (has blank errors) before adding it.


richmolj:
Yeah this looks solid, thanks so much for the fix @KirchoKirev ❤️ !

PR #358, opened 2021-05-08
Remove pagination links from show/find requests

### LINKED ISSUE #356: Pagination links are present on #show endpoints
When `Graphiti.config.pagination_links = true`

marcelobarreto:
Closes #356 

Just added a verification that ensure that the pagination links won't appear if it's a show action.

### DIFF
--- a/spec/query_spec.rb
+++ b/spec/query_spec.rb
@@ -994,4 +994,44 @@ RSpec.describe Graphiti::Query do
       end
     end
   end
+
+  describe '#pagination_links?' do
+    subject { instance.pagination_links? }
+
+    context 'when pagination_links_on_demand' do
+      before { Graphiti.config.pagination_links_on_demand = true }
+
+      context 'when params ask for pagination' do
+        let(:params) { { pagination_links: true } }
+
+        it { is_expected.to eq(true) }
+      end
+
+      context 'when params dont ask pagination' do
+        it { is_expected.to eq(false) }
+      end
+    end
+
+    context 'when action is equal to find' do
+      let(:params) { { action: 'show' } }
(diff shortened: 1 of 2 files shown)


spec/query_spec.rb
```diff
@@ -994,4 +994,44 @@
       end
     end
   end
+
+  describe '#pagination_links?' do
+    subject { instance.pagination_links? }
+
+    context 'when pagination_links_on_demand' do
+      before { Graphiti.config.pagination_links_on_demand = true }
```
richmolj: Since the config acts as a singleton, we need to make sure to reset this. Otherwise it can leak into other unrelated tests. Something like this (you'll see elsewhere):

```ruby
around do |e|
  original = Graphiti.config.pagination_links_on_demand
  Graphiti.config.pagination_links_on_demand = true
  begin
    e.run
  ensure
    Graphiti.config.pagination_links_on_demand = original
  end
end
```

richmolj (commented):
Thanks so much @marcelobarreto ❤️  I really appreciate the help. I can assure you this fixes a pain point for real-world users who will be thankful 👏 !

Left a tiny comment on the specs but otherwise LGTM 👍

spec/query_spec.rb
marcelobarreto: Yeah, that makes sense. Updated.

marcelobarreto: Is it okay now @richmolj? Sorry for this mistake, I saw that this was a singleton, but when I start to write the specs I completely forgot 😂

richmolj: No worries! Looks great! Thanks so much for the help ❤️

PR #361, opened 2021-05-16
Support page[offset] param

richmolj:
Say you want to start paging, 10 per page, but skip the first 8 results. Previously we didn't support this since the page count isn't a multiple of the offset - now we can! This will serve not only as a useful param in-and-of-itself, but also serve as the basis for future cursor pagination, which will default to offset-based (as opposed to stable-id-based) by default.

Small notes on backwards compatibility. If you have a custom `paginate` override we can't change the existing order so it's:

```ruby
paginate do |scope, current_page, per_page, context, offset|
```

Instead of `offset, context`.

Also, we check the adapter arity to see if we should pass `offset` - this should accomodate existing non-AR adapters.

### DIFF
--- a/lib/graphiti/adapters/abstract.rb
+++ b/lib/graphiti/adapters/abstract.rb
@@ -244,14 +244,15 @@ module Graphiti
       # @param scope The scope object we are chaining
       # @param [Integer] current_page The current page number
       # @param [Integer] per_page The number of results per page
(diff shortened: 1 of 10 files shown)


marcelobarreto (approved):

wadetandy (approved):
Looks great 👍

PR #363, opened 2021-06-08
Don't blow up on invalid JSON

richmolj:
We have a little helper that allows passing and parsing JSON strings. But that will blow up if passed invalid JSON, and the most common time that happens is when *the client isn't trying to pass json*. Consider:

```
?filter[name][eq]="[Setup]"
```

Prior to this commit, this request would blow up. But we shouldn't - the user is just asking for a string that has brackets in it.

This is only true if the type is `:string` - if the type is `:array` then yeah, we should blow up because they are trying to pass JSON and doing so incorrectly.


richmolj:
@wadetandy I consider this a bugfix rather than breaking compatibility, but would like to get your thoughts

wadetandy (approved):
Yeah this makes sense to me, both as a sensible change and as "bugfix vs enhancement". 👍

PR #373, opened 2021-08-08
Support cursor-based pagination

richmolj:
We currently support a `page[offset]` parameter. This updates the codebase to accept `before` and `after` cursors, applying the relevant `offset` logic as appropriate.

When JSON:API we're render `meta: { cursor: <cursor> }` for each entity, and when flat JSON/GraphQL we'll render `_cursor: <cursor>`. This is to support GQL queries like

```graphql
employees {
  nodes {
    firstName
    _cursor
  }
}
```

We're not doing a full `edges` implementation because A) I just kinda hate it and B) to value here is metadata about the relationship, mostly for M2M relationships, which Graphiti doesn't support yet anyway.

The cursor is Base64 encoded JSON of `{offset: <offset>}`.

In addition, we accept `?fields[page_info]=has_next_page,etc` to render the relevant `pageInfo` section. This uses the same code already used to populate pagination links and cursors.

### DIFF
--- a/lib/graphiti/errors.rb
+++ b/lib/graphiti/errors.rb
@@ -733,6 +733,12 @@ module Graphiti
       end
     end
 
+    class UnsupportedBeforeCursor < Base
+      def message
+        "Passing in page[before] and page[number] is not supported. Please create an issue if you need it!"
+      end
+    end
+
     class InvalidInclude < Base
       def initialize(resource, relationship)
         @resource = resource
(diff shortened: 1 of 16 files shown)


lib/graphiti/errors.rb
```diff
@@ -733,6 +733,12 @@ def message
       end
     end
 
+    class UnsupportedBeforeCursor < Base
+      def message
+        "Passing in page[before] and page[number] is not supported. Please create an issue if you need it!"
```
wadetandy, line 738: 👍

wadetandy (approved):

PR #374, opened 2021-08-09
Trim error messages

PChambino:
While writing a test on our application I noticed that the error messages include a lot of spaces at the beginning.

Since Graphiti supports ruby >= 2.3 I think we can use the squiggly heredoc to remove the leading spaces.

This may be a breaking change, but can also be considered patch if the leading spaces are not intentional.


richmolj:
I think we can call this a patch, thanks @PChambino ❤️ !

PR #379, opened 2021-08-23
Support schema for filters of extra attributes

### LINKED ISSUE #378: Filter of extra attribute fails when generating the schema
After adding a filter for an extra attribute, for example:
```ruby
extra_attribute :my_attr, :string, filterable: true
filter :my_attr, only: [:eq]
```

The `GraphitiSpecHelpers::RSpec.schema!` fails with:
```
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # /usr/local/bundle/gems/graphiti-1.2.41/lib/graphiti/schema.rb:200:in `block (2 levels) in filters'
```

Looking at the code ([schema.rb#L200](https://github.com/graphiti-api/graphiti/blob/7cac44258c292bc1a12c76cff41f5802f638ee8e/lib/graphiti/schema.rb#L200)) we can see that the schema generation doesn't consider extra attributes when generating filters. Also, the attributes in `resource.extra_attributes` don't have any `:schema` key by default. Feels like an oversight, but also not exactly sure what a good solution is for this specific case.

I guess I would have expected the filter of extra attributes to still be included in the generated schema.
The alternative would be to not include these filters in the schema for now so at least it doesn't fail with an error.

PChambino:
Defining a filter for an extra attribute causes the schema generation to
fail since it doesn't consider extra attributes.

This change allows filters (and sorts) of extra attributes to be
included in the generated schema. It also adds support for the `schema`
attribute option for extra attributes for parity with attributes.

Fixes https://github.com/graphiti-api/graphiti/issues/378

What do you think? For parity with attributes it makes sense to me,
the functionality is already there, this is makes schema generation work.

### DIFF
--- a/lib/graphiti/resource/dsl.rb
+++ b/lib/graphiti/resource/dsl.rb
@@ -151,7 +151,8 @@ module Graphiti
             readable: true,
             writable: false,
             sortable: false,
-            filterable: false
+            filterable: false,
+            schema: true
           }
           options = defaults.merge(options)
           attribute_option(options, :readable)
--- a/lib/graphiti/schema.rb
+++ b/lib/graphiti/schema.rb
@@ -153,6 +153,8 @@ module Graphiti
     def extra_attributes(resource)
       {}.tap do |attrs|
         resource.extra_attributes.each_pair do |name, config|
+          next unless config[:schema]
+
           attrs[name] = {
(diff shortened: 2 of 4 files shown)


richmolj:
Sorry for the delay @PChambino really appreciate the PR ❤️  Released in `1.3.5`

PR #380, opened 2021-08-23
Added an analogue ActiveRecord adapter for Sequel

AlexeyMatskevich:
Hi, this is my attempt to add an adapter for Sequel with functionality similar to ActiveRecord.
From the disadvantages of the adapter I will note several items:
  1. No support for aliases, I haven't found analog of `attribute_alias` for Sequel
  2. I don't fully understand what `#associate_all` and `#associate does` for `ActiveRecord` adapter, maybe my implementation doesn't quite fit as an analogue.
  3. `Kaminari` doesn't support `Sequel`, for that reason I used an implementation from the `Sequel` [extension](https://github.com/jeremyevans/sequel/blob/master/lib/sequel/extensions/pagination.rb) that doesn't support analog of [padding](https://github.com/kaminari/kaminari/blob/29d627dab1f51cf60cff65402c90229b1caa4891/kaminari-core/lib/kaminari/models/page_scope_methods.rb#L27).
  4. Perhaps Sequel `#disconnect` does not do the same as ActiveRecord `#clear_active_connections!`

### DIFF
--- a/lib/graphiti.rb
+++ b/lib/graphiti.rb
@@ -176,6 +176,10 @@ if defined?(ActiveRecord)
   require "graphiti/adapters/active_record"
 end
 
+if defined?(Sequel)
+  require "graphiti/adapters/sequel"
+end
+
 if defined?(Rails)
   require "graphiti/rails"
   require "graphiti/responders"
--- a/lib/graphiti/adapters/sequel.rb
+++ b/lib/graphiti/adapters/sequel.rb
@@ -1,0 +1,279 @@
+module Graphiti
+  module Adapters
+    class Sequel < ::Graphiti::Adapters::Abstract
+      require "graphiti/adapters/sequel/inference"
+      require "graphiti/adapters/sequel/has_many_sideload"
+      require "graphiti/adapters/sequel/belongs_to_sideload"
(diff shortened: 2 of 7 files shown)


richmolj:
Hey @AlexeyMatskevich sorry for the delay, I've had 👀  on this for a while but haven't commented. Great work. Very, very cool to see ❤️ 

Re: `associate/associate_all` - basically only there because AR will hit the DB as a side effect when using plain accessor logic, you might now need anything here.

Past that, I think this probably makes the most sense as a separate gem. I'd be happy to put it in the `graphiti-api` org and/or link to it on the documentation website, though.

Specs are the last thing. It'd be nice to have a kind of "adapter spec suite" we could run to validate an adapter works. Too bad we don't have that right now, but some sort of testing would be good.

Again, this was great to see!

AlexeyMatskevich:
Hey @richmolj I could try writing specifications for an ActiveRecord adapter to use this test set afterwards to compare the capabilities of the adapter for Seqeul with the standard adapter (ActiveRecord).
Then wee can take a look at the overall code and put the tests for AR in a separate pull request and the code of adapter for Sequel with tests in a separate gem.

PR #386, opened 2021-09-04
Handle malformed parameters with a 400 instead of a 500

sideshowbandana:
When processing an invalid request that is missing the following
headers:
```
'Content-Type': 'application/json'
```
and has the following body:

```
{'data': 'relationships'}
```

results in:

```
TypeError (no implicit conversion of Symbol into Integer):

activesupport (6.1.4.1) lib/active_support/core_ext/object/try.rb:15:in `[]'
activesupport (6.1.4.1) lib/active_support/core_ext/object/try.rb:15:in `public_send'
activesupport (6.1.4.1) lib/active_support/core_ext/object/try.rb:15:in `try'
graphiti (1.3.3) lib/graphiti/resource/links.rb:75:in `block in
allow_request?'
```
Which responds with a 500 status.

Instead the behavior should respond with a 400.

### DIFF
--- a/lib/graphiti/resource/links.rb
+++ b/lib/graphiti/resource/links.rb
@@ -72,14 +72,18 @@ module Graphiti
         request_path = request_path.split(".")[0]
 
         endpoints.any? do |e|
-          has_id = params[:id] || params[:data].try(:[], :id)
-          path = request_path
-          if [:update, :show, :destroy].include?(context_namespace) && has_id
-            path = request_path.split("/")
-            path.pop
-            path = path.join("/")
+          begin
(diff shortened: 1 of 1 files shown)


sideshowbandana:
@richmolj what do you think of this PR?

richmolj:
I think the issue is this breaks backwards-compatibility. I suggest throwing errors in this case and users can register those errors if they want 400. Also there is a codepath for this...you see we hit a `RequestValidator` for writes but it is a pass-through for reads...so making a true validator class is the way to go I think

sideshowbandana:
Sounds good, thanks for the suggestion!

PR #389, opened 2021-09-14
Fix Sparse fieldsets to work with relationships

### LINKED ISSUE #301: `fields` parameter is not considering relationship names as values
The `fields` parameter in `jsonapi-serializable` respects the relationship names(https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/serializer.rb#L35). So on `GET posts` api, if I request `fields[posts]=comments`, the response payload will only have `comments` relationship on posts. In graphiti this behaviour is overwritten here  https://github.com/jsonapi-rb/jsonapi-serializable/blob/master/lib/jsonapi/serializable/resource.rb#L95  and the `comments` in `fields`  parameter is not considered. 
With `include` param we are controlling data for which relationship is returned and not which relationship is rendered.
As per json api specs, fields can be attributes and relationships https://jsonapi.org/format/#document-resource-object-fields .

mrhardikjoshi:
Fixes #301

The sparse fieldsets is not filtering relationships. Currently if we specify any attribute and/or relationship in `fields` param, it works correctly on attributes but the not on relationships. It returns all the relationships, ignoring `fields` param. According to [jsonapi format](https://jsonapi.org/format/#fetching-sparse-fieldsets) both attribute and relationship should be supported. This PR fixes that.

In case of `fields` param having only attribute(s) in them, this change will restrict all relationships as well because of the following line from jsonapi format doc (fields represent both attributes and relationships)

> If a client requests a restricted set of fields for a given resource type, an endpoint MUST NOT include additional fields in resource objects of that type in its response.

I have not changed `include` section because of the following specification from the jsonapi format doc.
> The only exception to the full linkage requirement is when relationship fields that would otherwise contain linkage data are excluded via sparse fieldsets

So in case of having a relationship mentioned in `include` and sparse fieldsets having some atttribute, this will still return the relationship but the linkage won't be there.

### DIFF
--- a/lib/graphiti/serializer.rb
+++ b/lib/graphiti/serializer.rb
@@ -40,7 +40,7 @@ module Graphiti
         # conditional without affecting the relationship.
         def requested_relationships(fields)
           @_relationships.select do |k, _|
-            _conditionally_included?(self.class.relationship_condition_blocks, k)
+            _conditionally_included?(self.class.relationship_condition_blocks, k) && (fields.nil? || fields.include?(k))
           end
         end
       end
--- a/spec/fields_spec.rb
+++ b/spec/fields_spec.rb
@@ -2,7 +2,13 @@ require "spec_helper"
 
 RSpec.describe "fields" do
   include_context "resource testing"
-  let(:resource) { Class.new(PORO::EmployeeResource) }
+  let(:resource) do
+    Class.new(PORO::EmployeeResource) do
+      def self.name
+        "PORO::EmployeeResource"
+      end
+    end
+  end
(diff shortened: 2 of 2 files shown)


smai-f:
Just being the messenger from the Graphiti Slack so future perusers are aware...

image

https://graphiti-api.slack.com/archives/CJLJFRQFK/p1633956166007200?thread_ts=1632930343.006800&cid=CJLJFRQFK

jkeen:
I'm getting up to speed here and cleaning up some things and am going to have to trust the wisdom of @richmolj and close this PR until I'm more convinced that this is a real issue.

PR #391, opened 2021-10-19
Use eql when filtering polymorphic_has_many or polymorphic_has_one 

timflapper:
Hi, Polymorphic associations in ActiveRecord are being filtered using the `lower` database function on the `:#{as}_type` attribute. This causes indices that include the polymorphic association attributes to be ignored. By forcing `eql` on the filter, the `:#{as}_type` attribute will no longer be queried with the `lower` function, which solves the issue.


timflapper:
Hi @richmolj, do you have any feedback on this PR?

richmolj (approved):
Good change!

richmolj:
Thanks ❤️ !

timkrins:
I came across an interesting incompatibility with this particular change when bumping versions today.

It looks like a symptom of the way we were defining our polymorphic relationships -
In some resources which we are sideloading, we were specifying some of our `polymorphicable_type` fields as a `string_enum`, rather than a `string`, and passing a list of known valid polymorphic type values.

This now causes an error when sideloading, as the `string_enum` type does not have the correct `eql` filter.

I might switch all these fields in our app to just be a `string` type, which would solve my particular problem, ~just need to figure out potential impact of allowing any string rather than the defined list of class names~.

Some generified error strings for others who might be searching for the same issue:
`Graphiti::Errors::SideloadQueryBuildingError`
`error occurred while sideloading`
`Graphiti::Errors::UnsupportedOperator`
`Tried to filter :polymorphicable_type on operator :eql, but not supported! Supported operators are [:eq].`

Update: turns out my understanding of the difference (and functionality) of `string_enum` was wrong, and a `string` field also supports the `allow` argument (and even when you are using an `allow` field with a `string_enum`, this is only enforced when filtering, and **not** when persisting).

Replacing all the `string_enum` attributes with `string` for the polymorphic types fixes my issue.

PR #400, opened 2022-01-26
Drop EOLed Rubies and Rails on CI

wagenet:


richmolj:
You're the best Peter ❤️ !

PR #404, opened 2022-02-03
feat: active_record adapter #filter_array_eq and #filter_array_not_eq

mileszim:


jkeen:
@mileszim Could you describe the use-case here and add potentially add some tests for it?

jkeen:
@mileszim Also, in looking at this again—should the order matter? Doesn't look like that compare code is sorting the values

PR #421, opened 2022-06-15
Re-use resource class for remote sideloads (avoid memory leak)

PChambino:
Dynamically creating resource classes for each sideload is expensive on
allocations and it was causing a memory leak on carwow's production
applications since these Class objects were not being freed.

Not sure if there was anything in particular to our applications that
was preventing these Class objects from being freed, but even if there
is no memory leak on a simple application, it is still expensive and slow
in terms of allocations.

When running a [memory-profiler](https://github.com/SamSaffron/memory_profiler) and doing an HTTP requests that
includes/sideloads resources from a remote resource, you should be able
to see something like:

```
retained memory by location
-----------------------------------
   2.23 kB  /usr/local/bundle/gems/graphiti-1.3.6/lib/graphiti/query.rb:81
```

This PR proposes defining a static resource class for remote sideloads
which can be re-used: `RemoteSideloadResource`.


richmolj (approved):
Makes total sense, thanks ❤️ !

PR #423, opened 2022-06-28
Prefer #update over #update_attributes

jkeen:
Using/allowing `@resource.update` instead of `@resource.update_attributes` might help out a lot of rubocop + graphiti users, as Rubocop loves to think that `@resource.update_attributes` is out-of-date Rails code and will automatically change it to `@resource.update` which before this update broke things. 

This keeps graphiti a little more in sync with the standard Railsy controller pattern

### DIFF
--- a/lib/graphiti/resource_proxy.rb
+++ b/lib/graphiti/resource_proxy.rb
@@ -153,10 +153,12 @@ module Graphiti
       success
     end
 
-    def update_attributes
+    def update
       data
       save(action: :update)
     end
+
+    alias update_attributes update
 
     def include_hash
       @include_hash ||= begin


jkeen:
Note: `Base64` and `Position` related test failures are addressed in #422

richmolj:
Thanks Jeff LGTM but could you give some context on the spec failures? Were they failing prior to this PR?

jkeen:
@richmolj They were, they're actually failing in a bunch of the open PRs. It's completely unrelated to this change. While prepping this I noticed them and PR'd that fix early. 

If you can approve and merge #422 a bunch of the failures will get resolved

PR #429, opened 2022-07-20
boolean_attribute sets type as :boolean

jasonkarns:
I encountered this by reading the code, not encountering any particular issue.

Shouldn't `boolean_attribute` be setting the attribute type to `:boolean` automatically?


jasonkarns:
apologies, i got confused thinking this was from the Resource, not the serializer.

PR #431, opened 2022-08-16
fix: leverage ruby-2.7 parameter forwarding

belt:
... to allow ruby-3.0 positional and named parameter splitting to
work within resources


belt:
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

PR #433, opened 2022-09-02
polymorphic `on` expects a symbol

jasonkarns:
Polymorphic belongs-to expects a symbol name as the "type" for a given subclass. In our case, we have had to customize Rails' polymorphic type derivation logic. So for any given AR model, we can't simply use the model's name in our polymorphic belongs-to.

To prevent duplicating (and decoupling) the knowledge of what a given model's polymorphic type name is, we're using:

```rb
  polymorphic_belongs_to :approvable do
    group_by(:approvable_type) do
      on(Address.polymorphic_name) # is actually AddressList
```

However! This fails to match because the grouping logic is matching against `:AddressList` as a symbol.

We could of course .to_sym ourselves, but I think if graphiti intends to match against a symbol it should do the conversion itself to ensure the matching value is also a symbol.


jkeen:
@jasonkarns Mind adding a test for this?

PR #442, opened 2022-12-20
Fix nested validations

carlobeltrame:
Nested validation errors on update operations are currently broken.
ActiveRecord objects are compared for equality by id only. The |= operator would ignore any children which are only updated, i.e. whose id has not changed. By reversing the `target = target | [child]` logic to `target = [child] | target`, we elect to keep the child with its updated attributes. This is relevant for detecting validation errors in the child later.

Let me know if you generally approve of this change, so I can start investing time into writing tests for it.


jkeen:
@carlobeltrame I just hopped in the ring to help here and am getting up to speed and sorting through these PRs and issues

I haven't personally run into this problem and I'm not sure if I'm fully understanding what's going on, but getting the tests running and new ones added to illustrate the issue you're solving would be great.

PR #446, opened 2023-01-24
Fix private call, it had the wrong argument

daniel-illi:
This should make method `relationship_option` private instead `relationship_option` as it follows the `def relationship_option` method and the other `def relationship_option` was already made private before.


jkeen:
Good catch

PR #470, opened 2024-03-19
Add thread pool and concurrency_max_threads configuration option

### LINKED ISSUE #469: ActiveRecord::ConnectionTimeoutError "all pooled connections were in use" after enabling concurrency
I recently enabled concurrency on a system and it resulted in a lot of `ActiveRecord::ConnectionTimeoutError` errors.

I believe it's because ActiveRecord has a 1 thread = 1 connection policy and the database connection pool for this system is limited to the number of puma threads. As soon as the system receives enough requests to use up the connection pool, the concurrently side-loaded resource threads are unable to get a connection from the pool, and vice versa.

But I'm surprised no one has come across this issue before so perhaps I've misunderstood the problem.

One solution might be to increase the system's database connection pool size. This would allow Graphiti's threads to have enough connections to not exhaust the thread pool. The difficulty would be in sizing the db pool. Given each request can sideload `x` resources, the total number of database connections required at any given time would be `x * number of concurrent requests`.

Ideally though I think it would make sense to allow the thread pool size to be configured. `Concurrent::Promise`, used to side-loaded resources, has by default a thread pool equal to java's int max.

What do you think?

```
vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb:128:in `block in wait_poll': could not obtain a connection from the pool within 5.000 seconds (waited 5.005 seconds); all pooled connections were in use (ActiveRecord::ConnectionTimeoutError)
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb:117:in `loop'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb:117:in `wait_poll'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb:78:in `internal_poll'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb:204:in `internal_poll'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb:73:in `block in poll'
    from vendor/ruby-3.2.2/lib/ruby/3.2.0/monitor.rb:202:in `synchronize'
    from vendor/ruby-3.2.2/lib/ruby/3.2.0/monitor.rb:202:in `mon_synchronize'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb:82:in `synchronize'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb:73:in `poll'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool.rb:663:in `acquire_connection'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool.rb:353:in `checkout'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_pool.rb:182:in `connection'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_adapters/abstract/connection_handler.rb:246:in `retrieve_connection'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_handling.rb:287:in `retrieve_connection'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/connection_handling.rb:254:in `connection'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/querying.rb:62:in `_query_by_sql'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/relation.rb:998:in `block in exec_main_query'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/relation.rb:1018:in `skip_query_cache_if_necessary'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/relation.rb:984:in `exec_main_query'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/relation.rb:962:in `block in exec_queries'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/relation.rb:1018:in `skip_query_cache_if_necessary'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/relation.rb:956:in `exec_queries'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/relation.rb:742:in `load'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/relation.rb:264:in `records'
    from vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.3/lib/active_record/relation.rb:259:in `to_ary'
    from lib/graphiti_adapters/an_adapter.rb:19:in `resolve'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti/resource.rb:117:in `resolve'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti/scope.rb:22:in `block in resolve'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti/scope.rb:81:in `block in broadcast_data'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti.rb:65:in `block in broadcast'
    from vendor/bundle/ruby/3.2.0/gems/activesupport-7.1.3/lib/active_support/notifications.rb:206:in `block in instrument'
    from vendor/bundle/ruby/3.2.0/gems/activesupport-7.1.3/lib/active_support/notifications/instrumenter.rb:58:in `instrument'
    from vendor/bundle/ruby/3.2.0/gems/activesupport-7.1.3/lib/active_support/notifications.rb:206:in `instrument'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti.rb:64:in `broadcast'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti/scope.rb:80:in `broadcast_data'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti/scope.rb:20:in `resolve'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti/resource_proxy.rb:68:in `data'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti/sideload.rb:231:in `load'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti/sideload.rb:299:in `resolve'
    from vendor/bundle/ruby/3.2.0/gems/graphiti-1.3.9/lib/graphiti/scope.rb:47:in `block (2 levels) in resolve_sideloads'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/safe_task_executor.rb:24:in `block in execute'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/synchronization/mutex_lockable_object.rb:48:in `block in synchronize'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/synchronization/mutex_lockable_object.rb:48:in `synchronize'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/synchronization/mutex_lockable_object.rb:48:in `synchronize'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/safe_task_executor.rb:22:in `execute'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/promise.rb:564:in `block in realize'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:359:in `run_task'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:350:in `block (3 levels) in create_worker'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:341:in `loop'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:341:in `block (2 levels) in create_worker'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:340:in `catch'
    from vendor/bundle/ruby/3.2.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:340:in `block in create_worker'
```

MattFenelon:
This option allows to limit the maximum number of resources that can be
sideloaded concurrently. With a properly configured connection pool,
this ensures that the activerecord's connection pool is not exhausted by
the sideloading process.

The thread pool configuration is based on ActiveRecord's
global_thread_pool_async_query_executor.

Closes #469

See: https://github.com/rails/rails/blob/94faed4dd4c915829afc2698e055a0265d9b5773/activerecord/lib/active_record.rb#L279-L287

### DIFF
--- a/lib/graphiti/configuration.rb
+++ b/lib/graphiti/configuration.rb
@@ -7,8 +7,20 @@ module Graphiti
     # @return [Boolean] Concurrently fetch sideloads?
     #   Defaults to false OR if classes are cached (Rails-only)
     attr_accessor :concurrency
+    
+    # This number must be considered in accordance with the database
+    # connection pool size configured in `database.yml`. The connection
+    # pool should be large enough to accommodate both the foreground
+    # threads (ie. web server or job worker threads) and background
+    # threads. For each process, Graphiti will create one global
+    # executor that uses this many threads to sideload resources
+    # asynchronously. Thus, the pool size should be at least
+    # `thread_count + concurrency_max_threads + 1`. For example, if your
+    # web server has a maximum of 3 threads, and
+    # `concurrency_max_threads` is set to 4, then your pool size should
+    # be at least 8.
     # @return [Integer] Maximum number of threads to use when fetching sideloads concurrently
-    attr_accessor :concurrency_pool_max_size
+    attr_accessor :concurrency_max_threads
 
     attr_accessor :respond_to
     attr_accessor :context_for_endpoint
@@ -28,7 +40,7 @@ module Graphiti
     def initialize
       @raise_on_missing_sideload = true
       @concurrency = false
-      @concurrency_pool_max_size = 4
+      @concurrency_max_threads = 4
       @respond_to = [:json, :jsonapi, :xml]
       @links_on_demand = false
       @pagination_links_on_demand = false
--- a/lib/graphiti/configuration.rb
+++ b/lib/graphiti/configuration.rb
@@ -7,6 +7,8 @@ module Graphiti
     # @return [Boolean] Concurrently fetch sideloads?
     #   Defaults to false OR if classes are cached (Rails-only)
     attr_accessor :concurrency
+    # @return [Integer] Maximum number of threads to use when fetching sideloads concurrently
+    attr_accessor :concurrency_pool_max_size
 
     attr_accessor :respond_to
     attr_accessor :context_for_endpoint
@@ -26,6 +28,7 @@ module Graphiti
     def initialize
       @raise_on_missing_sideload = true
       @concurrency = false
+      @concurrency_pool_max_size = 4
       @respond_to = [:json, :jsonapi, :xml]
       @links_on_demand = false
       @pagination_links_on_demand = false
--- a/lib/graphiti/configuration.rb
+++ b/lib/graphiti/configuration.rb
@@ -7,7 +7,7 @@ module Graphiti
     # @return [Boolean] Concurrently fetch sideloads?
     #   Defaults to false OR if classes are cached (Rails-only)
     attr_accessor :concurrency
-    
+
     # This number must be considered in accordance with the database
     # connection pool size configured in `database.yml`. The connection
     # pool should be large enough to accommodate both the foreground
--- a/lib/graphiti/scope.rb
+++ b/lib/graphiti/scope.rb
@@ -25,20 +25,20 @@ module Graphiti
       @query = query
       @opts = opts
 
-      @object = @resource.around_scoping(@object, @query.hash) do |scope|
+      @object = @resource.around_scoping(@object, @query.hash) { |scope|
         apply_scoping(scope, opts)
-      end
+      }
     end
 
     def resolve
       if @query.zero_results?
         []
       else
-        resolved = broadcast_data do |payload|
+        resolved = broadcast_data { |payload|
           @object = @resource.before_resolve(@object, @query)
           payload[:results] = @resource.resolve(@object)
           payload[:results]
-        end
+        }
         resolved.compact!
         assign_serializer(resolved)
         yield resolved if block_given?
--- a/lib/graphiti/scope.rb
+++ b/lib/graphiti/scope.rb
@@ -3,14 +3,20 @@ module Graphiti
     attr_accessor :object, :unpaginated_object
     attr_reader :pagination
 
+    @thread_pool_executor_mutex = Mutex.new
+
     def self.thread_pool_executor
+      return @thread_pool_executor if @thread_pool_executor
+
       concurrency = Graphiti.config.concurrency_max_threads || 4
-      @thread_pool_executor ||= Concurrent::ThreadPoolExecutor.new(
-        min_threads: 0,
-        max_threads: concurrency,
-        max_queue: concurrency * 4,
-        fallback_policy: :caller_runs
-      )
+      @thread_pool_executor_mutex.synchronize do
+        @thread_pool_executor ||= Concurrent::ThreadPoolExecutor.new(
+          min_threads: 0,
+          max_threads: concurrency,
+          max_queue: concurrency * 4,
+          fallback_policy: :caller_runs
+        )
+      end
     end
 
     def initialize(object, resource, query, opts = {})
@@ -19,20 +25,20 @@ module Graphiti
       @query = query
       @opts = opts
 
-      @object = @resource.around_scoping(@object, @query.hash) { |scope|
+      @object = @resource.around_scoping(@object, @query.hash) do |scope|
         apply_scoping(scope, opts)
-      }
+      end
     end
 
     def resolve
       if @query.zero_results?
         []
       else
-        resolved = broadcast_data { |payload|
+        resolved = broadcast_data do |payload|
           @object = @resource.before_resolve(@object, @query)
           payload[:results] = @resource.resolve(@object)
           payload[:results]
-        }
+        end
         resolved.compact!
         assign_serializer(resolved)
         yield resolved if block_given?
--- a/lib/graphiti/scope.rb
+++ b/lib/graphiti/scope.rb
@@ -2,6 +2,17 @@ module Graphiti
   class Scope
     attr_accessor :object, :unpaginated_object
     attr_reader :pagination
+
+    def self.thread_pool_executor
+      concurrency = Graphiti.config.concurrency_pool_max_size || 4
+      @thread_pool_executor ||= Concurrent::ThreadPoolExecutor.new(
+        min_threads: 0,
+        max_threads: concurrency,
+        max_queue: concurrency * 4,
+        fallback_policy: :caller_runs
+      )
+    end
+
     def initialize(object, resource, query, opts = {})
       @object = object
       @resource = resource
@@ -49,7 +60,7 @@ module Graphiti
           @resource.adapter.close if concurrent
         }
         if concurrent
-          promises << Concurrent::Promise.execute(&resolve_sideload)
+          promises << Concurrent::Promise.execute(executor: self.class.thread_pool_executor, &resolve_sideload)
         else
           resolve_sideload.call
         end
--- a/lib/graphiti/scope.rb
+++ b/lib/graphiti/scope.rb
@@ -4,7 +4,7 @@ module Graphiti
     attr_reader :pagination
 
     def self.thread_pool_executor
-      concurrency = Graphiti.config.concurrency_pool_max_size || 4
+      concurrency = Graphiti.config.concurrency_max_threads || 4
       @thread_pool_executor ||= Concurrent::ThreadPoolExecutor.new(
         min_threads: 0,
         max_threads: concurrency,
--- a/spec/configuration_spec.rb
+++ b/spec/configuration_spec.rb
@@ -150,18 +150,18 @@ RSpec.describe Graphiti::Configuration do
     end
   end
 
-  describe "#concurrency_pool_max_size" do
-    include_context "with config", :concurrency_pool_max_size
+  describe '#concurrency_max_threads' do
+    include_context "with config", :concurrency_max_threads
 
     it "defaults" do
-      expect(Graphiti.config.concurrency_pool_max_size).to eq(4)
+      expect(Graphiti.config.concurrency_max_threads).to eq(4)
     end
 
     it "is overridable" do
       Graphiti.configure do |c|
-        c.concurrency_pool_max_size = 1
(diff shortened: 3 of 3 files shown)


MattFenelon:
Is it possible the test failure is flaky? It doesn't look related to the changes.

jkeen:
@MattFenelon yep, I re-ran it and the tests pass now. 

I've run into some thread pool issues on the app I use graphiti with, but never traced it back to the sideloads… so this is an interesting find!

It looks like the new code sets the default at 4? I'm not seeing where that default was set before your change, and just want to ensure that this doesn't do anything unexpected for people who are running just fine. In your configuration you set the value at 1?

MattFenelon:
> @MattFenelon yep, I re-ran it and the tests pass now.

Nice. Thank you!

> It looks like the new code sets the default at 4? I'm not seeing where that default was set before your change, and just want to ensure that this doesn't do anything unexpected for people who are running just fine. In your configuration you set the value at 1?

That's a little tricky to answer :) The short answer is that the code was using Concurrent::Promise and that uses concurrent-ruby's global io thread pool, which has a max size equal to Java's max int: https://github.com/ruby-concurrency/concurrent-ruby/blob/8b9b0da4a37585ce5eb71516aca55e93bde39115/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb#L15.

The problem with the default is that people will likely see these threading issues pop up so I think the default needs to be set lower. However, lowering the default will mean that people may see less concurrency happening than they were before. I went with 4 as that's what Rails' uses as its default pool size for async queries. But it's fairly arbitrary.

I have tested it configured to 1 and that works happily because the fallback option `:caller_runs` just runs on the calling thread. That also means no one should see any errors from the change because the calling thread will just be used to run anything that goes beyond the queue/pool size. That's why I didn't want to set the default to 1 because the concurrency will be limited to 1 thread / 1 sideload.

Perhaps it would be worth updating the docs to tell people they need to set their db connection pool appropriately otherwise even with this change they may see threading issues.

wdyt?

jkeen:
@MattFenelon Thanks for the details.

> The problem with the default is that people will likely see these threading issues pop up so I think the default needs to be set lower. However, lowering the default will mean that people may see less concurrency happening than they were before. I went with 4 as that's what Rails' uses as its default pool size for async queries. But it's fairly arbitrary.

But your theory is that if they're seeing those errors with the new explicit default set at 4, they'd have been seeing them with the implicit default being at 4 currently?

I only recently became a maintainer on this repo and don't yet have access to editing docs. @richmolj's time is totally consumed by a tsunami of life so I'm not sure when he'll get me plugged into that side of things.  https://www.graphiti.dev/guides/concepts/resources#concurrency

If I'm understanding the issue correctly though, will having `concurrency = true` (default in rails) and the new `concurrency_pool_max_size` option set to `1`, is that effectively the same as having `concurrency = false` ?

MattFenelon:
> If I'm understanding the issue correctly though, will having `concurrency = true` (default in rails) and the new `concurrency_pool_max_size` option set to `1`, is that effectively the same as having `concurrency = false` ?

Sorry, I could have been clearer. `1` will set the pool size to 1 so there'll be 1 thread in the pool to be used for side loading. The max queue size would be 1 * 4 so any side loads above that will just be loaded using the calling thread.

> If you want to provide a maximum queue size, you may also consider the fallback_policy which defines what will happen if work is posted to a pool when the queue of waiting work has reached the maximum size and no new threads can be created. Available policies: 
> ...
> caller_runs: The work will be executed in the thread of the caller, instead of being given to another thread in the pool.
> **From https://ruby-concurrency.github.io/concurrent-ruby/master/file.thread_pools.html**

> But your theory is that if they're seeing those errors with the new explicit default set at 4, they'd have been seeing them with the implicit default being at 4 currently?

That's it exactly. There's basically no safe default without configuring the database pool correctly. This just lowers the possibility as there'll only ever be 4 threads side loading at once rather than `DEFAULT_MAX_POOL_SIZE`. But I think setting concurrency to false by default to be safer is probably a more surprising change at this point. What do you think? 

We could set the default to 1 but even so there's still a chance the pool will become exhausted. For example, with puma set to 5 threads and the pool set to 5, if the system is processing 5 requests at the same time the connection pool is already exhausted and sideloading any resources at that point will raise the error. But `1` does lower the risk at the expense of less performance. I'm not sure what the right balance is.

jkeen:
I think your logic is sound? I'm not sure what the best default is that would hit that mark of best performance without issues. I feel tiny bit out of my depth when it comes to thread-related issues, but one thing that jumps out at me is that if this new setting is just for concurrency relating to sideloads, I think naming the configuration option to more clearly reflect that would be helpful. `Graphiti.config.sideload_max_threads`, perhaps?

MattFenelon:
> I think naming the configuration option to more clearly reflect that would be helpful

Sounds good. I went with `concurrency_max_threads` just because the `concurrency` setting already says it's for sideloads and I think it's nice that the settings sit next together alphabetically.

https://github.com/graphiti-api/graphiti/blob/f68b61ff09ec61ecf23acc5bc37d0accba14aeed/lib/graphiti/configuration.rb#L7-L9

Thanks for the reviews on this!

MattFenelon:
I just realised that the executor initialisation wasn't thread-safe so I've fixed that.

jkeen:
@MattFenelon Oh didn't realize that re: the concurrency flag. Your naming makes perfect sense then—nice work, and good catch on the thread-safety issue.

I fixed a silly lint error only the CI robots care about, but the rest looks good to me and I'll cut a minor release with this change. Thanks for the contribution!

MattFenelon:
> I fixed a silly lint error only the CI robots care about, but the rest looks good to me and I'll cut a minor release with this change. Thanks for the contribution!

Ha thanks. My editor really wants to auto format a lot of that code 😅

Looking forward to the release!

jkeen:
@MattFenelon I had to make a small change to the thread-safety logic (1.6.1) after running into some major lockups in production (as in, the web server would stop responding) with the change as it was. When trying it out locally, the change I made to how the mutex was set up made all the difference, but redeploying this just now still causes a lock-up in production when concurrency is enabled? I don't know what/why that is happening? I'm at the end of a long day here, but if you see this in the morning see if anything jumps out at you?

MattFenelon:
Right ok. I think it might be because the mutex variable is not actually global but an instance variable on the class. It needs to be global to work properly. We could make it a constant or a class variable ‘@@mutex’. I just can’t remember all of the inheritance issues with constants and class variables off the top of my head.I’m not at a computer right now but I can raise a PR in a bit.

MattFenelon:
@jkeen I've raised a new PR to fix the issue. I can't say for certain that it was what caused the issue you saw but it seems likely.

MattFenelon:
Thanks again for the merge @jkeen. Apologies for the muck up there.

jkeen:
@MattFenelon No problem at all, thanks for the quick turnaround!

PR #471, opened 2024-03-22
Fix thread pool scope

MattFenelon:
The thread pool and mutex need to be global across instances of Scope for it to be a global thread pool.

See: https://github.com/graphiti-api/graphiti/pull/470


jkeen:
references this issue: https://github.com/graphiti-api/graphiti/issues/469

PR #472, opened 2024-03-27
Add thread pool with promises to limit concurrent sideloading

### LINKED ISSUE #469: ActiveRecord::ConnectionTimeoutError "all pooled connections were in use" after enabling concurrency

MattFenelon:
Previously, sideloading in Graphiti used an unbounded number of threads,
which could decrease performance due to resource contention and exhaust the
ActiveRecord connection pool, leading to timeouts and connection errors.

This commit introduces a thread pool based on ActiveRecord's
`global_thread_pool_async_query_executor`, limiting the maximum number of
resources that can be sideloaded concurrently. With a properly configured
connection pool, this ensures that the ActiveRecord connection pool is not
exhausted by the sideloading process.

Promises are used to manage recursive links between threads, avoiding
deadlocks that were reported in earlier attempts.

Single-threaded mode is reintroduced with a new implementation that uses
promises in both async and sync modes. The difference lies in whether the
thread pool is set to synchronous or not.

Parent thread's local variables are copied to the sideloading threads,
ensuring that any thread locals available to the Graphiti resolving thread
are also accessible to the sideload threads. Similarly, parent fiber's local
variables are copied to the sideloading threads, preserving fiber-local
context during sideloading.

Thread pool statistics broadcasting is added for observability. Collecting
stats on the thread pool can be useful to understand the state of the thread
pool, for example, if many tasks are being queued.

Bump `concurrent-ruby` gem from version 1.0 to 1.2+, which includes a fix
for worker threads getting stuck when using the `caller_runs` policy
(ruby-concurrency/concurrent-ruby#933).

### TODO
- [x] Fix deadlock
- [x] Fix rails integration
- [x] Refactor `ResourceProxy#data` to another method for concurrent
- [x] Does save break, i.e. does `resolve_sideloads` return a promise?
- [x] Close adapter
- [x] Thread locals
- [x] Thread pool stats broadcast (to be tested)

Closes https://github.com/graphiti-api/graphiti/issues/469

See: https://github.com/rails/rails/blob/94faed4dd4c915829afc2698e055a0265d9b5773/activerecord/lib/active_record.rb#L279-L287
See: https://github.com/graphiti-api/graphiti/pull/470

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request: {}
 
 concurrency:
@@ -14,7 +14,7 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: "3.0"
@@ -94,7 +94,7 @@ jobs:
       - name: Set up Appraisal
         if: matrix.appraisal
         run: echo "APPRAISAL_INITIALIZED=true" >> $GITHUB_ENV
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 graphiti changelog
 
+## [1.7.6](https://github.com/graphiti-api/graphiti/compare/v1.7.5...v1.7.6) (2024-11-06)
+
+
+### Bug Fixes
+
+* Gem version check ([#483](https://github.com/graphiti-api/graphiti/issues/483)) ([68e2492](https://github.com/graphiti-api/graphiti/commit/68e2492032692d8bb928a733f8b0f8710be31c49))
+
 ## [1.7.5](https://github.com/graphiti-api/graphiti/compare/v1.7.4...v1.7.5) (2024-09-16)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 graphiti changelog
 
+## [1.7.4](https://github.com/graphiti-api/graphiti/compare/v1.7.3...v1.7.4) (2024-09-11)
+
+
+### Bug Fixes
+
+* update ActiveRecord adapter w/ support for Rails 7.2+ ([#478](https://github.com/graphiti-api/graphiti/issues/478)) ([8313e33](https://github.com/graphiti-api/graphiti/commit/8313e3359f0dde28d9940867c7ded964db4c854d))
+
 ## [1.7.3](https://github.com/graphiti-api/graphiti/compare/v1.7.2...v1.7.3) (2024-06-26)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 graphiti changelog
 
+## [1.7.1](https://github.com/graphiti-api/graphiti/compare/v1.7.0...v1.7.1) (2024-04-18)
+
+
+### Bug Fixes
+
+* properly display .find vs .all in debugger statements ([d2a7a03](https://github.com/graphiti-api/graphiti/commit/d2a7a038a649818979d52ccd898e68dba78b051f))
+* rescue error from sideloads updated_at calculation, defaulting to the current time ([661e3b5](https://github.com/graphiti-api/graphiti/commit/661e3b5212e2649870a200067d0d5d52fa962637))
+
 # [1.7.0](https://github.com/graphiti-api/graphiti/compare/v1.6.4...v1.7.0) (2024-03-27)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 graphiti changelog
 
+# [1.7.0](https://github.com/graphiti-api/graphiti/compare/v1.6.4...v1.7.0) (2024-03-27)
+
+
+### Features
+
+* Add support for caching renders in Graphiti, and better support using etags and stale? in the controller ([#424](https://github.com/graphiti-api/graphiti/issues/424)) ([8bae50a](https://github.com/graphiti-api/graphiti/commit/8bae50ab82559e2644d506e16a4f715effd89317))
+
+## [1.6.4](https://github.com/graphiti-api/graphiti/compare/v1.6.3...v1.6.4) (2024-03-27)
+
 ## [1.6.3](https://github.com/graphiti-api/graphiti/compare/v1.6.2...v1.6.3) (2024-03-26)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 graphiti changelog
 
+## [1.7.5](https://github.com/graphiti-api/graphiti/compare/v1.7.4...v1.7.5) (2024-09-16)
+
+
+### Bug Fixes
+
+* Fixes error in version check for ActiveRecord adapter introduced in [#478](https://github.com/graphiti-api/graphiti/issues/478) ([#479](https://github.com/graphiti-api/graphiti/issues/479)) ([42c82c3](https://github.com/graphiti-api/graphiti/commit/42c82c397f20eb91c02835e518ff4c351c028ea7))
+
 ## [1.7.4](https://github.com/graphiti-api/graphiti/compare/v1.7.3...v1.7.4) (2024-09-11)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 graphiti changelog
 
+## [1.7.3](https://github.com/graphiti-api/graphiti/compare/v1.7.2...v1.7.3) (2024-06-26)
+
+
+### Bug Fixes
+
+* require OpenStruct explicitly ([#475](https://github.com/graphiti-api/graphiti/issues/475)) ([e0fa18a](https://github.com/graphiti-api/graphiti/commit/e0fa18a8d7f051e385e6e081f79f2ecae92a9260))
+
+## [1.7.2](https://github.com/graphiti-api/graphiti/compare/v1.7.1...v1.7.2) (2024-06-11)
+
+
+### Bug Fixes
+
+* require necessary ActiveSupport parts in proper order ([bb2a488](https://github.com/graphiti-api/graphiti/commit/bb2a48874a6533522df6eb027d0df8ec14c80a20))
+
 ## [1.7.1](https://github.com/graphiti-api/graphiti/compare/v1.7.0...v1.7.1) (2024-04-18)
 
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,13 +1,22 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
 # Specify your gem's dependencies in graphiti.gemspec
 gemspec
 
+github 'rails/rails' do
+  gem 'activemodel'
+  gem 'rails'
+end
+
+gem 'database_cleaner'
+gem 'graphiti-rails', '~> 0.4.0'
+gem 'rspec-rails'
+gem 'sqlite3'
+
 group :test do
-  gem "database_cleaner"
-  gem "pry"
-  gem "pry-byebug", platform: [:mri]
-  gem "appraisal"
-  gem "guard"
-  gem "guard-rspec"
+  gem 'appraisal'
+  gem 'guard'
+  gem 'guard-rspec'
+  gem 'pry'
+  gem 'pry-byebug', platform: [:mri]
 end
--- a/Gemfile
+++ b/Gemfile
@@ -1,22 +1,22 @@
-source 'https://rubygems.org'
+source "https://rubygems.org"
 
 # Specify your gem's dependencies in graphiti.gemspec
 gemspec
 
-github 'rails/rails' do
-  gem 'activemodel'
-  gem 'rails'
+github "rails/rails" do
+  gem "activemodel"
+  gem "rails"
 end
 
-gem 'database_cleaner'
-gem 'graphiti-rails', '~> 0.4.0'
-gem 'rspec-rails'
-gem 'sqlite3'
+gem "database_cleaner"
+gem "graphiti-rails", "~> 0.4.0"
+gem "rspec-rails"
+gem "sqlite3"
 
 group :test do
-  gem 'appraisal'
-  gem 'guard'
-  gem 'guard-rspec'
-  gem 'pry'
-  gem 'pry-byebug', platform: [:mri]
+  gem "appraisal"
+  gem "guard"
+  gem "guard-rspec"
+  gem "pry"
+  gem "pry-byebug", platform: [:mri]
 end
--- a/Gemfile
+++ b/Gemfile
@@ -3,20 +3,11 @@ source "https://rubygems.org"
 # Specify your gem's dependencies in graphiti.gemspec
 gemspec
 
-github "rails/rails" do
-  gem "activemodel"
-  gem "rails"
-end
-
-gem "database_cleaner"
-gem "graphiti-rails", "~> 0.4.0"
-gem "rspec-rails"
-gem "sqlite3"
-
 group :test do
+  gem "database_cleaner"
+  gem "pry"
+  gem "pry-byebug", platform: [:mri]
   gem "appraisal"
   gem "guard"
   gem "guard-rspec"
-  gem "pry"
-  gem "pry-byebug", platform: [:mri]
 end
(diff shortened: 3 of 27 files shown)


jkeen:
Still getting a lockup. This sounded possibly related: https://github.com/ruby-concurrency/concurrent-ruby/issues/933

And it looks like graphiti only requires `concurrency-ruby ~> 1.0`, when really it should be ` > 1.2` to include that fix?  However, I tried including that and trying it out and it didn't seem to help.

But one thing i realized: one request that's causing a hang has 13 sideloads… which is a lot, admittedly. But still, if there are more sideloads than there are threads, I think that might be a trouble spot?

MattFenelon:
> Still getting a lockup. This sounded possibly related: [ruby-concurrency/concurrent-ruby#933](https://github.com/ruby-concurrency/concurrent-ruby/issues/933)
> 
> And it looks like graphiti only requires `concurrency-ruby ~> 1.0`, when really it should be ` > 1.2` to include that fix? However, I tried including that and trying it out and it didn't seem to help.
> 
> But one thing i realized: one request that's causing a hang has 13 sideloads… which is a lot, admittedly. But still, if there are more sideloads than there are threads, I think that might be a trouble spot?

Thanks for testing that, @jkeen. Good find on that bug. Upgrading the gem is definitely something to consider but as you say, it doesn't seem to have solved the issue.

I'm surprised we're having issues as the code is very similar to that used by ActiveRecord for the new async queries. So perhaps there's something in the way that Graphiti is calling the thread pool that is causing problems. I suspect that's more the case then there being a bug in concurrent-ruby's thread pool.

13 sideloads seems a lot but the default thread pool has 4 threads and a queue of 4 * 4 that (25) and if the queue is expired it's meant to use the calling thread to run the request. I'm wondering if that the :caller_runs option might be where the problem lies.

Are all of the sideloads running ActiveRecord queries?

I think I might need to add some logging to debug the issue.

jkeen:
All the sideloads run active record queries, and some are pulling nested relationships off the sideloads, as in `includes=items,items.comments`, etc.

I think figuring out a good logging solution would be great improvement—I dug around the ruby-concurrency docs briefly and couldn't make heads or tails of how one might log when a new thread is getting created or hung up. But if we had some good logs this issue might be much easier to diagnose.

As far as the other options for fallback, I haven't tried :abort or :drop. I would assume abort would make the request throw a 500, and drop would… return incomplete data?

MattFenelon:
I've added a failing test case that shows the deadlock:

https://github.com/graphiti-api/graphiti/pull/472/commits/e4de93e5108ddefeff8d3f9639a1ca950dccbef8

jkeen:
Excellent!! Nice work on that 🥇

MattFenelon:
@jkeen do you know how to get the rails tests to run locally? I'm struggling to debug the failures there.

I've tried running:

    $ bundle exec appraisal install

Which fails when trying to install the sqlite3 gem

`An error occurred while installing sqlite3 (1.4.4), and Bundler cannot continue.`

The installed sqlite3 is the default that comes with macOS

    $ sqlite3 --version
    3.32.2 2021-07-12 15:00:17 bcd014c473794b09f61fbc0f4d9488365b023f16123b278dbbd49948c27calt2

I've also tried using a later version.

jkeen:
@MattFenelon Sqlite is notorious for throwing build errors and I feel like I've spent countless hours of my life troubleshooting system-specific compilation errors with sqlite with ruby and node projects, and with the time in that space I should be an expert, and yet every time it comes along I'm no better equipped than I was the last time.

Did you try installing the gem separately first? This has also helped me in the past: https://github.com/sparklemotion/sqlite3-ruby/blob/main/INSTALLATION.md

MattFenelon:
> @MattFenelon Sqlite is notorious for throwing build errors and I feel like I've spent countless hours of my life troubleshooting system-specific compilation errors with sqlite with ruby and node projects, and with the time in that space I should be an expert, and yet every time it comes along I'm no better equipped than I was the last time.
> 
> Did you try installing the gem separately first? This has also helped me in the past: https://github.com/sparklemotion/sqlite3-ruby/blob/main/INSTALLATION.md

Sounds about right 😅

I've managed to work around it by upgrading to version 2 of the sqlite3 gem as that uses native binaries. I also had to upgrade rails to the main branch version (8) but that's got me to a point where I can debug the rails tests.

jkeen:
@MattFenelon Are you pretty satisfied where this work is at now? I haven't tried it again in my staging or production app, just wanted to get your read on it before I give it another spin.

MattFenelon:
> @MattFenelon Are you pretty satisfied where this work is at now? I haven't tried it again in my staging or production app, just wanted to get your read on it before I give it another spin.

I was but there seems to be some flakiness for some of the tests that could be related to the changes but I'm not sure. Do you know if those tests are already flaky?

I have tested the changes locally with a test server with various configurations of rails thread pool, graphiti thread pool and the db connection pool. I've not had a chance to test in production. If you think the test is a red herring, I'd be really interested to hear your experiences of testing it.

I would like to finish this. I plan to come back to it but it probably won't be for a few weeks now.

MattFenelon:
@jkeen I'm very close with this now and hope to be able to open the PR for review next week. I have had the patch running in production in two services for more than a week with no problems. The last thing I would like to check is that the ActiveSupport::Notification I added for the thread pool works as expected. I'm hoping to be able to test that this week.

jkeen:
@MattFenelon Great news! I've been following along and it looks like things have progressed a bunch. Looking forward to testing this out whenever you're ready. Is the notifications/global_thread_pool_stats part of the puzzle for letting someone know if configurations need to be changed? 

The thread pool numbers calculation bit of everything I hope we can make as simple and foolproof as possible

MattFenelon:
@jkeen This is ready now! I'm interested in hearing your thoughts.

> Is the notifications/global_thread_pool_stats part of the puzzle for letting someone know if configurations need to be changed?

Exactly! The thread pool is a point of contention. All sideloads across _all requests_ go through the thread pool. For instance, with the default configuration of 4 threads, if one request is using up all 4 threads to sideload, the next request will have to wait until one of the threads is available to sideload its own resources. It's not exactly as clear cut as that because threads free up all of the time but it gives a simple overview.

It's a trade-off but it ensures the safety of the system by ensuring that it will never use up more database connections than there are available.

With the thread pool stats notification, the state of the thread pool can be observed at run time in order to tune to each system's exact needs. If `graphiti.thread_pool_queue_length` is showing that tasks are sitting in the queue often, that's perhaps a sign that an extra thread or two would be beneficial.

That's for advanced use cases though. The default of 4 seems like a reasonable starting point for most applications, particularly as it's the default for Rails' load_async thread pool.

MattFenelon:
@jkeen I was thinking more about the impact this change will have once merged. It might be surprising for users of the gem to find the use of threads changed. How about I configure the thread pool to work the same way as before, i.e. the threads would be unbounded? I could perhaps also add a deprecated flag to signify the problems with setting concurrency without concurrency_max_threads. What do you think?

jkeen:
@MattFenelon If it's the right thing to do ultimately (and it seems like it is given the hidden problems you explained about the current method) I'm fine with it being a change, bumping the minor version. If you think the new way is going to break stuff unexpectedly then we could bump the major version even. 

The one thing I don't want to do is have some great improvement that requires someone to find a hidden option or read the documentation in order to find and benefit from; the best way should be the default—and it seems like what you've been working on is the best way. Would you agree?

MattFenelon:
> ...If you think the new way is going to break stuff unexpectedly then we could bump the major version even.

A minor bump feels ok to me. Internally it's a decent size change but it shouldn't break anything or require any changes on the part of consumers of the gem.

> The one thing I don't want to do is have some great improvement that requires someone to find a hidden option or read the documentation in order to find and benefit from; the best way should be the default—and it seems like what you've been working on is the best way. Would you agree?

Yes I agree. This change makes the default configuration safer by fixing #469. It's also likely to have improve performance. With the current set up of an unbounded number of threads, there can be a high contention for resources, which particularly affects the latency of the long-tail of requests. In my testing on my local system, I found even a small increase of one thread to the thread pool (4 to 5) decreased the overall requests per second the system could handle.

It sounds like we're in agreement so, if you're happy with the change, I think this PR is good to go.

PR #477, opened 2024-08-28
Fix deprecated call in activerecord adapter

nbarthel:
Deprecated call `::ActiveRecord::Base.clear_active_connections!` has been moved to connection_handler.

This was causing our app to fail in production on Rails 7.2.1.

Not sure how well 7.2.1 is supported yet.  Perhaps more PRs to come.

### DIFF
--- a/lib/graphiti/adapters/active_record.rb
+++ b/lib/graphiti/adapters/active_record.rb
@@ -304,7 +304,7 @@ module Graphiti
       end
 
       def close
-        ::ActiveRecord::Base.clear_active_connections!
+        ::ActiveRecord::Base.connection_handler.clear_active_connections!
       end
 
       def can_group?


lib/graphiti/adapters/active_record.rb
```diff
@@ -304,7 +304,7 @@ def destroy(model_instance)
       end
 
       def close
-        ::ActiveRecord::Base.clear_active_connections!
+        ::ActiveRecord::Base.connection_handler.clear_active_connections!
```
mihaimuntenas, line 307: if ActiveRecord.version > '7.2'
  ::ActiveRecord::Base.connection_handler.clear_active_connections!
else
  ::ActiveRecord::Base.clear_active_connections!
end

mihaimuntenas: Let's assume that not everyone would be able to upgrade to 7.2+ as fast as we need to do it ;)

mihaimuntenas: as a quick fix, in the meantime, we could  use something like the following:
`ActiveRecord::Base.class.delegate :clear_active_connections!, to: :connection_handler`
somewhere in an initializer, or such.

mihaimuntenas: or a much cleaner version:
```
ActiveRecord::Base.define_singleton_method :clear_active_connections! do
  connection_handler.clear_active_connections!
end
```

jkeen:
@nbarthel Mind making that change and updating the PR? I'll release a new version after that.

I like the clarity of the first version @mihaimuntenas suggested.

mihaimuntenas:
@jkeen I opened #478 to speed things up.

jkeen:
Fixed in #478

PR #483, opened 2024-11-05
Fix gem version check

ef4:
All the other places in graphiti that compare against `ActiveRecord.version` are doing it this way for compatibility with new-enough dependencies. This one seems to have been missed.


jkeen:
thanks @ef4!

PR #485, opened 2024-11-11
Add field condition block class attribute to avoid error in ruby 3.2+

jkeen:


jkeen:
This was a rails issue and was resolved in a patch version of Rails 8

PR #496, opened 2025-05-15
WIP: Maintain thread context in spawned threads for nested sideloads

jkeen:
When there are multi-level sideloads, the thread context is not maintained when sideloads spawn sideloads, i.e. when a sideload of "permissions.user" is requested. In practice this makes the context object available in resource methods (used to access controller methods) unavailable.

In my app this was causing intermittent 500 errors when it couldn't access the controller to grab the pundit context. 
@MattFenelon Removing these lines fixes the issue, but I'm not sure if this will cause other unintended problems?

### DIFF
--- a/lib/graphiti/scope.rb
+++ b/lib/graphiti/scope.rb
@@ -163,10 +163,10 @@ module Graphiti
             yield(*args)
           end
 
-          if execution_context_changed
-            thread_storage&.keys&.each { |key| Thread.current[key] = nil }
-            fiber_storage&.keys&.each { |key| Fiber[key] = nil }
-          end
+          # if execution_context_changed
+          #   thread_storage&.keys&.each { |key| Thread.current[key] = nil }
+          #   fiber_storage&.keys&.each { |key| Fiber[key] = nil }
+          # end
 
           result
         end


MattFenelon:
I think that may cause memory leaks as any data added in a thread won’t be cleaned down at the end of the thread. I can’t think why removing those lines fixes the issue you’re seeing as the context should be set back up at the start of a new thread (execution context).

jkeen:
@MattFenelon It only seems to happen when the includes are nested and it looked like when I was debugging that in my example of my request with a handful of sideloads plus `permissions.user`, the `permissions` sideload thread would execute, then another sideload would execute and then a `user` sideload would execute with the parent resource being permissions…but the thread context would be lost.

MattFenelon:
I wonder if it’s because the execution context hasn’t changed, causing the if statement body to not execute, which would copy the context to Thread.current again. In other words,

1. Thread 1 runs
2. Sets context
3. Loads scope
4. Clears context
5. Thread 1 runs next scope
6. Context hasn’t changed so doesn’t set it again
7. Loads scope
8. Fails because the code relying on the context causes an exception.

jkeen:
Hmmm, this sounds like a possibility. Any idea how to handle that?

MattFenelon:
I think there'd need to be two changes. The first is to always set the thread variables at the start of the work and remove the `if execution_context_changed` conditional. That will fix the issue you're seeing of the disappearing context.

The second thing is to fix the clearing of the thread/fiber storage, before more work is scheduled on that thread, while removing `execution_context_changed` (as it doesn't work). I think what we want here is to compare what was in the thread when the promise was started and only remove (set to nil) those keys that we set at the start of the promise, not those that were already there. We should probably move that to an `ensure` as well.

What do you think?

jkeen:
@MattFenelon Yeah that sounds like a good plan

MattFenelon:
I've had a go at fixing it over in https://github.com/graphiti-api/graphiti/pull/497, @jkeen.

jkeen:
Closed in favor of #497

PR #497, opened 2025-05-20
Fix lost thread and fiber locals

MattFenelon:
Thread and fiber locals are being cleared on completion of every future. If the promise ran on the same thread as the request, or the same thread as the last promise, the locals are then not available. The execution context hasn't changed in those circumstances so the code that sets the locals at the start of a promise doesn't run.

1. Thread 1 runs
2. Sets context
3. Loads scope
4. Clears context
5. Thread 1 runs next scope
6. Context hasn’t changed so doesn’t set it again
7. Loads scope
8. Fails because the code relying on the context causes an exception because the locals aren't available.

Instead always set the thread and fiber locals and only clear those that we know are new. This should avoid the lost locals and also avoid memory leaks between promises.

See: https://github.com/graphiti-api/graphiti/pull/496

### DIFF
--- a/lib/graphiti/scope.rb
+++ b/lib/graphiti/scope.rb
@@ -176,6 +176,8 @@ module Graphiti
     end
 
     def with_fiber_locals(fiber_locals)
+      return yield unless fiber_locals
+
       new_fiber_locals = []
       fiber_locals.each do |key, value|
         if !Fiber[key]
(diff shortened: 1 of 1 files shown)


jkeen:
Thanks @MattFenelon! This looks/works great

MattFenelon:
Good to hear! Thanks for the merge, @jkeen.

PR #498, opened 2025-07-29
Add cache tags support

factyy:
The original cache implementation relies on two main parts: scope and query.

Quite often we use scoping as a policy-based access control for listing actions like `index`. With Graphiti it is easily achieved by making `#base_scope` context-aware (e.g. like checking a requesting user role and restricting access to certain entities for non-admins). The problem is that it is only represented via changes in the DB query object (we have different cache keys for different scopes since they result in different queries). One of the side effects is that Graphiti has no idea about it and cache debugger stubbornly states lots and lots of cache misses (it associates cache keys with the request info that doesn't include any info from the request context). 

The idea behind this PR is simple: enable cache tagging with a value returned from a resource instance method (that can access request context). This method name has to be passed as a symbol argument to the `cache_resource` resource class method alongside `expires_in` argument.

### DIFF
--- a/lib/graphiti/resource/interface.rb
+++ b/lib/graphiti/resource/interface.rb
@@ -4,9 +4,10 @@ module Graphiti
       extend ActiveSupport::Concern
 
       class_methods do
-        def cache_resource(expires_in: false)
+        def cache_resource(expires_in: false, tag: nil)
           @cache_resource = true
           @cache_expires_in = expires_in
+          @cache_tag = tag
         end
 
         def all(params = {}, base_scope = nil)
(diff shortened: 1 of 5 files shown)


superslau:
I have a very similar issue, so interested to see what is necessary for this to get merged.

factyy:
@jkeen , sorry to bother, is there anything I can help you with to merge this PR?

jkeen:
@factyy Sorry for the delay, I think this is good? I'm not sure why that one test is failing though, and it looks like I left this sit for too long and I can't seem to re-run it on CI to make sure it's just an intermittent thing

factyy:
@jkeen , probably it is still good, as far as I remember there weren't any breaking changes...

PR #500, opened 2025-12-30
Support rails 8.1

bguban:
Looke like rails 8.1 removed ActionDispatch::Journey::Router::Utils.uri_decoder method and now the default way to decode the url is to use URI.decode_uri_component


jkeen:
@bguban Thanks for the PR—looks like there are a bunch of failing tests?

bguban:
@jkeen, yeah, but looks like they are not related to my changes. I created a different PR where I changed only README.md https://github.com/graphiti-api/graphiti/pull/501 and it also has a bunch of failed tests.

bguban:
I added tests for rails 8.1 just to be sure that nothing else is broken

jkeen:
Thanks for adding the readme change test, that clarifies some things. The 8.x failures make sense on the readme test, and there's a single 7.x failure that seems to have appeared since last release due to an upstream change I'm guessing, but this change for rails 8 seems to cause a bunch of failures in the 5, 6, and 7 appraisals that all seem to be related to that `url_decoder` thing that are not present in the readme test 🤔

bguban:
Oh, my bad, I accidentally checked for the wrong method. Looks like everything is good now.

PR #502, opened 2026-01-09
Make parse_fieldset more resilient

amaierhofer:
Querys with `?fields[model]=` are already handled okay but when omitting `=` it results in an exception. This changes fixes that


jkeen:
Nice fix!

PR #506, opened 2026-07-08
Adds attr_reader to Graphiti::Errors::UnsupportedOperator

mapreal19:
We want to build our own error detail message without the resource class name `#message` includes

We could do it now via `instance_variable_get` but that's brittle across future gem releases


mapreal19:
@jkeen thank you for taking over 🙏

PR #515, opened 2026-07-31
fix: typecast null values in filter lists

### LINKED ISSUE #411: Using null filter as part of comma delineated filter list does not typecast 'null' to nil
If filtering by nil is enabled on your resource and you use a filter such as `filter[name]=null,Lucy`, it would be nice if this filter returned all values that were either nil or Lucy. Instead, the filter value ends up [here](https://github.com/graphiti-api/graphiti/blob/master/lib/graphiti/adapters/active_record.rb#L41) as `['null', 'Lucy']` -- `'null'` is a string and does not get typecast to `nil`.

Only verified with resource backed by ActiveRecord.

lntutor:
## Summary

- typecast `null` entries inside comma-delimited filter values when `allow_nil` is enabled
- preserve existing behavior for filters that do not allow nil values
- add a regression spec covering `filter[name]=null,Lucy`

Fixes #411

## Verification

- `bundle exec rspec spec/filtering_spec.rb`
- `bundle exec rspec`
- `bundle exec standardrb`
- `git diff --check`

### DIFF
--- a/lib/graphiti/scoping/filter.rb
+++ b/lib/graphiti/scoping/filter.rb
@@ -209,7 +209,9 @@ module Graphiti
     end
 
     def parse_string_null(filter, value)
-      return if value == "null" && filter[:allow_nil]
+      return value unless filter[:allow_nil]
+      return value.map { |item| item == "null" ? nil : item } if value.is_a?(Array)
+      return if value == "null"
 
       value
     end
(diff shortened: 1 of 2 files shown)


lntutor:
Current head `4a4a7a07bdaad2ccce344fa9bb9bd6bce544675f` is mergeable and the visible CI surface on this exact SHA is fully green: `Lint` plus the full Ruby/Appraisal matrix across supported Rails combinations all completed successfully. There are no review threads or review-body comments open on the current PR state, so if this null-filter typecast fix still matches the intended direction, it looks ready for maintainer review/merge when convenient.

PR #516, opened 2026-08-01
Chore/release pipeline

caseyprovost:


jkeen:
Dropping Ruby 2.x support and Rails 5.x/6.x support is a breaking change/major version release, and I partially started that process on the `beta` branch along with another long-awaited breaking change. I think all this stuff also belongs there? It won't be too long before I cut a graphiti 2.0 release, and I'm cutting beta gem releases as I add features.

Wanna repoint this PR to beta?

PR #537, opened 2026-08-08
fix: reject invalid page parameters

### LINKED ISSUE #347: NoMethodError when page is not a hash
Request:

```
/some_controller?limit=10&page=1
```

Error:

```
undefined method `each_pair' for "1":String
Did you mean?  each_char
```

The error comes from here:

```
# graphiti/query.rb
 def pagination
      @pagination ||= begin
        {}.tap do |hash|
          (@params[:page] || {}).each_pair do |name, value|
```

Would you accept a pull request to handle that gracefully?

aouxwoux:
closes #347

`page=1` currently reaches `Query#pagination`, which assumes `page` responds to `each_pair` and raises a `NoMethodError`.

this validates `page` at the existing request-validator boundary before read requests bypass payload validation. non-object values now produce `Graphiti::Errors::InvalidRequest` with `page must be an object`, while omitted and hash-shaped page parameters keep their existing behavior.

checks:
- `bundle exec rspec spec/request_validator_spec.rb spec/query_spec.rb` (103 examples, 0 failures)
- `bundle exec standardrb lib/graphiti/request_validators/validator.rb spec/request_validator_spec.rb`
- `git diff --check`

### DIFF
--- a/lib/graphiti/request_validators/validator.rb
+++ b/lib/graphiti/request_validators/validator.rb
@@ -11,10 +11,12 @@ module Graphiti
       end
 
       def validate
+        validate_page
+
         # Right now, all requests - even reads - go through the validator
         # In the future these should have their own validation logic, but
         # for now we can just bypass
-        return true unless @params.has_key?(:data)
+        return errors.blank? unless @params.has_key?(:data)
 
         resource = @root_resource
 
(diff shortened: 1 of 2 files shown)


jkeen:
Thanks @aouxwoux!

PR #538, opened 2026-08-09
fix: treat empty polymorphic configuration as unset

### LINKED ISSUE #199: Consistent behaviour for empty polymorphic
This is just a small issue, but I would expect
```ruby
    self.polymorphic = []
```
to behave the same as no assignment at all. Instead, the behaviour changes, since the `Polymorphism` module is included (in `Graphiti::Resource::Configuration`)

alloutflo:
## Summary

- make an empty `polymorphic` collection report as non-polymorphic
- delegate already-prepended polymorphism hooks to normal resource behavior when the collection becomes empty
- preserve the unknown-child regression with a genuinely partial child list

This also handles the case raised in the issue discussion where a previously populated collection later becomes empty: the behavior checks the current collection rather than relying only on assignment time.

Fixes #199

## Validation

- `bundle exec rspec spec/polymorphism_spec.rb` — 18 examples, 0 failures
- `bundle exec rspec` — 1,421 examples, 0 failures, 21 pending
- `bundle exec standardrb` — passed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ spec/.rspec-examples
 spec/dummy/log/*
 *~
 .vscode/
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@ spec/.rspec-examples
 spec/dummy/log/*
 *~
 .vscode/
-.DS_Store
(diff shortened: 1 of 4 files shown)


jkeen:
Thanks @alloutflo!

PR #539, opened 2026-08-09
Graphiti 2.0  🚀

### LINKED ISSUE #167: Always include id and type for belongs_to relationships
Summarizing the lengthy discussion on [slack](https://graphiti-api.slack.com/archives/C5A4UEMGS/p1565624466070600):

Currently graphiti does not include the resource identifier object of a relationship without explicitly requesting the resource be included. This makes sense in a `has_many` case where there might be thousands of related records, but perhaps not in the case of a `belongs_to` where the foreign key is right there on the loaded resource.

Not rendering the relationship data on the resource when it is easily available can cause problems, such as when using a client side jsonapi library with good caching like ember-data. Currently if I've loaded some parent records in one request, and later I load child records directly without explicitly specifying to include the parent, the payload will be rendered without any relationship data (only `{meta: {included: false}}`, perhaps links) and the client will have no way to relate the two together without a) re-requesting the relationship alone (through its link) or b) specifying the parent be sideloaded with the child request with `?include=parent`. Both of these options require the server to re-render a resource that the client already has loaded, which feels inefficient.

Futher, it just feels weird that in a pre-jsonapi world a resource with a `belongs_to` would have had a `parent_id` attribute included in the payload, allowing the client to link things together that are already loaded, or even just to know which child records are grouped together without having to know anything about the parent.

`jsonapi-serializable` [includes an option](https://github.com/jsonapi-rb/jsonapi-serializable/issues/94#issuecomment-351097652) to specify `linkage: always: true` on a `belongs_to` to achieve this behavior, and I think graphiti should set this by default.

### LINKED ISSUE #199: Consistent behaviour for empty polymorphic

### LINKED ISSUE #216: Localization Support
Are there plans to add localization support to Graphiti? I'd like to be able to offer multilingual error messaging.

### LINKED ISSUE #347: NoMethodError when page is not a hash

### LINKED ISSUE #518: Resource Cheatsheet references 'ApplicationRecord' should be 'ApplicationResource'
I'm just reading the docs to start using the project and found this typo

https://github.com/graphiti-api/graphiti-api.github.io/blob/c98df4d27a28f4ff7434fad801e399eccfed66ab/_site/cheatsheet.html#L88

### LINKED ISSUE #520: Generated specs require `GraphitiSpecHelpers::Sugar`
We need it for the `d` method used here:
https://github.com/graphiti-api/graphiti/blob/master/lib/generators/graphiti/templates/index_request_spec.rb.erb

Relatedly, the docs say that `jsonapi_data` is aliased to `d` but AFAICT, `Sugar` is not mentioned. https://www.graphiti.dev/guides/concepts/testing#jsonapi-data

### LINKED ISSUE #524: Sideloads randomly stop getting included in responses
Have a bizarre issue where sideloads randomly stop getting included in responses periodically. It took me hours to debug this. Finally when I all else fails, I restart my server, and bam, they're back.

This is in `development` and I haven't actually tried in prod yet as I'm still vetting the library. 

Any ideas?

I am loading all my classes in development fyi, because it helps me spot some bugs. Maybe involved?

### LINKED ISSUE #526: Exception handling
Could someone provide a more detailed explanation of the way exception handling works? I added

```ruby
  register_exception ActiveRecord::RecordNotFound, status: 404
```
to my controller, but it does not seem to be caught. I stepped into it and verified that it is added to a list of handled exceptions, but couldn't find what I need to do for these to actually be rescued.

I previously used `rescue_from`, and don't mind using it again, but I would need to produce a standard jsonapi response, and couldn't find a method (in either `graphiti` or `graphiti-rails`) that would do that. Also couldn't find something that converts `object.errors` to such a format. I assume I'm missing something fundamental...

### LINKED ISSUE #528: ResourceGenerator overwrites existing controller
Overwriting existing controllers when generating a resource limits the utility of the gem. It makes it much more difficult to add Graphiti to an existing application.

**Pain point**
Adding Graphiti to an existing application with `generate graphiti:resource [RESOURCE_NAME] --model=[MODEL_NAME]` overwrites the existing controller. You can skip the overwrite, but then the controller code must be written by hand. Also, if you use `destroy graphiti:resource...` it deletes the controller file whether the generator initially overwrote the file or not. This inflexibility also increases the difficulty of calling the generator inside other custom generators.

**Suggested solutions:**
- Break out the responsibilities of the `ResourceGenerator` into separate generators and create a single scaffold generator which incorporates all of the generators
  - The scaffold generator could have options to exclude specific generators (e.g. `--no-controller`)
- Instead of overwriting controllers, create separate namespaced controllers specifically for Graphiti. The file structure could be sourced from the namespace recorded in `.graphiticfg.yml`.

### LINKED ISSUE #530: Auto-include `Responders` module if responders gem is present
This is something that might be worth doing. We should also consider any potential downsides.

### LINKED ISSUE #534: Error responses missing exception message details
Graphiti's error classes have a lot of `message` methods with helpful errors details. But as far as I can tell, graphiti-rails has the rendering of these disabled because it doesn't pass `detail: :exception` to `rescue-registry` here:

https://github.com/graphiti-api/graphiti-rails/blob/4a8a065f9c67cf0c3bcedfb7c8ac3da033560e57/lib/graphiti/rails.rb#L46-L49

This results in `detail: nil` in the jsonapi errors.

### LINKED ISSUE #535: Include Graphiti stuff only in api
It seems that the stuff here:
https://github.com/graphiti-api/graphiti-rails/blob/2b787402f3247a5fdb27e0cb36f53522dac49557/lib/graphiti/rails.rb#L39
is included in all controllers, not just the ones that use Graphiti. Is it possible to opt-out?

### LINKED ISSUE #540: Thread.current causes issues with fiber based servers like falcon
Hi,

currently i do these patches (AI generated).
``` 
Rails.application.config.after_initialize do
  # Patch 1: Graphiti.context (request-scoped metadata)
  module Graphiti
    class << self
      def context
        (Fiber.current.storage || {}).fetch(:graphiti_context, nil) || {}
      end

      def context=(val)
        verbose, $VERBOSE = $VERBOSE, nil
        Fiber.current.storage = (Fiber.current.storage || {}).merge(graphiti_context: val)
      ensure
        $VERBOSE = verbose
      end
    end
  end

  # Patch 2: TransactionHooksRecorder (before/after commit hooks)
  module Graphiti
    module Util
      class TransactionHooksRecorder
        class << self
          private

          def _hooks
            (Fiber.current.storage || {}).fetch(:_graphiti_hooks, nil)
          end

          def reset_hooks
            verbose, $VERBOSE = $VERBOSE, nil
            Fiber.current.storage = (Fiber.current.storage || {}).merge(
              _graphiti_hooks: {
                after_graph_persist: [],
                before_commit: [],
                after_commit: []
              }
            )
          ensure
            $VERBOSE = verbose
          end
        end
      end
    end
  end

  # Patch 3: RescueRegistry (exception handler context)
  module RescueRegistry
    class << self
      def context
        (Fiber.current.storage || {}).fetch(:rescue_registry_context, nil)
      end

      def context=(value)
        verbose, $VERBOSE = $VERBOSE, nil
        Fiber.current.storage = (Fiber.current.storage || {}).merge(rescue_registry_context: value)
      ensure
        $VERBOSE = verbose
      end
    end
  end
end
``` 

Sorry no tests that i can contribute. Would be nice to see graphiti working in modern web servers.

jkeen:
# Graphiti 2.0

Graphiti 2.0 collapses the satellite-gem constellation into one gem, makes persistence behave the way you always assumed it did, and ships with a rebuilt documentation site. There are some important breaking changes that improve ergonomics, but nearly everything from 1.x still runs under 2.0 with deprecation warnings, so most apps can upgrade incrementally. Upgrade path here: https://graphiti.dev/upgrading/

## One gem instead of four

`graphiti-rails`, `graphiti_spec_helpers`, and `graphiti_errors` are now part of graphiti itself. Remove them from your Gemfile after upgrading `graphiti` and everything keeps working. Rails integration is now opt-in per controller with `include Graphiti::Rails::Controller`, instead of every controller silently receiving it (#535). Exception handling is built on `rescue_registry` (#526), and spec helpers live at `Graphiti::SpecHelpers` with a new set of RSpec matchers for asserting resource attributes and relationships (graphiti-api/graphiti_spec_helpers#14).

This collapse was primarily motivated by the fact that 80-90% of graphiti users are using it with rails, and collapsing the surface area makes maintenance and understanding easier.

## Persistence that matches your mental model

The model you inspect is the model that saves (#465, [docs](https://graphiti.dev/concepts/persisting/)). Hooks and overrides see the same assigned model instance that gets persisted, instead of a parallel attributes hash that could drift from reality. `around_persistence` hooks receive that model, writable guards now receive the model and attribute name so authorization can consider the actual record (#511), and the resource carries its assigned model rather than threading it through override signatures.

Writable guards evaluate under the action being performed on every path, so `writable: proc { current_action == :create }` answers the same whether or not you inspected the model first.

## Serialization wins

- `belongs_to` renders relationship resource ids by default, solving an issue that had been open since 2019 (#167), with `belongs_to_resource_ids_by_default` and a per-relationship `resource_ids` option to control it. The default comes without a performance cost, since linkage renders from foreign keys already on the record.
- `Resource.wrap` serializes models you fetched yourself, so you can use graphiti's serialization without going through its finders (#513).
- A subclass redeclaring a relationship now reaches its own serializer.
- Sideloaded entities are deduplicated across include paths, so a record reached through two different associations renders once, complete. Previously only the first copy to serialize got its relationships, leaving `"data": null` holes in included (#476).
- A relationship linkage no longer disappears when an unrelated include happens to share its name. `?include=positions.classification` used to blank the employee's own classification.
- Deduplication no longer widens a deep filtered association. `?include=positions.employee.positions&filter[positions.rank]=1` returned every position where the shallow path correctly returned one.

## Concurrency

Concurrent sideload resolution is more reliable, with fixes for deadlocks and swallowed errors and new stress specs covering behavior under load. Request state now lives in fiber storage rather than thread locals, so context, transaction hooks and debug chunks survive into fibers a request spawns. That makes graphiti work under fiber based servers such as falcon (#540), and closes a case where a sideload resolving inside a fiber would wait on the bounded pool from a pool thread.

`ActiveSupport::CurrentAttributes` are carried onto pool threads, so a sideload resolving concurrently still sees `Current.user` and anything else your app sets per request. Previously that state was left behind on the request thread, and a resource reading it during a concurrent sideload saw nothing.

## Performance

Serialization is faster and allocates less, with the biggest gains on simple requests. Every release records its object allocations so changes are visible over time, and `rake performance:current` reports them locally.

## Guardrails and tooling

- `rake graphiti:audit` scans your resources for latent issues.
- `rails g graphiti:locale` writes every error code graphiti renders into a locale file, ready to customize or translate. 
- New rake tasks for schema checks, so verifying a schema no longer means running the whole suite.
- Rendering a relationship the model doesn't define now raises `MissingRelationshipMethod` instead of failing obscurely.
- The resource generator takes a `--controller` name, so it no longer overwrites a controller you already have (#528).
- `context_namespace` is now `current_action`, which says what it actually is.

## A documentation site that lives with the code

The docs site was rebuilt, absorbed into this repository, and now publishes automatically as new versions are cut. Runtime error messages link to the new docs, and there's a dedicated 2.0 upgrade guide. Behavior and documentation updates can ship at the same time!

## Modern platform

Requires Ruby 3.2+ and Rails 7.1+. Tested against Ruby 3.2 through 4.0 and Rails 7.1 through 8.1, on a rebuilt CI matrix and release pipeline. Apps on older Rubies and Rails stay supported on the 1.x maintenance branch.

## Breaking changes

The [upgrade guide](https://graphiti.dev/upgrading/) walks through each of these with migration steps. The short version:

- Ruby 3.2+ and Rails 7.1+ are required.
- Controllers serving graphiti resources must `include Graphiti::Rails::Controller`. Under graphiti-rails, every controller received the integration automatically (#535).
- `around_persistence` hooks receive the assigned model instead of the attributes hash (#465). Move attribute-hash edits to `before_attributes`, or set values on the model. Custom `create`/`update` overrides keep their 1.x signatures and need no changes.
- `include GraphitiErrors` raises instead of warning, because rescue_registry replaced it and there is nothing left to point it at. `GraphitiErrors.enable!`/`disable!` becomes `handle_request_exceptions`, and 409 responses now report code "conflict" and title "Conflict Error".
- `belongs_to` relationships render resource linkage by default, so those relationship objects gain a `data` key in payloads (#167). Set `belongs_to_resource_ids_by_default = false` for 1.x output.
- Relationships that render neither ids nor a link are omitted rather than rendered empty.
- `allow_nil` and `deny_empty` fold into one `blanks:` filter option, and the pagination settings are named after the page params they control.

Everything else from 1.x resolves with a deprecation warning and keeps working until 3.0.

## Closes

- Closes #167 - `belongs_to` renders resource linkage by default, from foreign keys already on the record so no extra queries
- Closes #518 - the rebuilt docs site fixed the cheatsheet typo
- Closes #520 - generated specs call the helpers by their full names, and the `Sugar` aliases are deprecated
- Closes #526 - one exception-handling stack via rescue_registry, `register_exception` behaves as documented
- Closes #528 - `--controller` lets a resource point at a controller without overwriting it
- Closes #530 - superseded, Rails integration is now explicit opt-in rather than auto-include
- Closes #535 - controllers opt in with `include Graphiti::Rails::Controller` (transferred from graphiti-rails#52)
- Closes #540 - request state moved to fiber storage, so graphiti works under falcon and other fiber based servers
- Closes #524 - `included` always renders when the client asked for it, so resolving the proxy first no longer drops sideloads
- Closes #534 - registered exceptions render their message as `detail`, and every code's text is overridable from a locale key
- Closes #216 - error titles and details come from locale keys, with a generator that writes every code graphiti renders
- Closes #347 - invalid page parameters are rejected rather than raising NoMethodError
- Closes #199 - an empty polymorphic configuration is treated as unset
- Closes #476 - duplicate sideloaded entities now squash to one complete object; @factyy's diagnosis and approach landed here directly, since the PR had drifted from the 2.0 internals
- Closes #516 - pipeline, dependency, and linter updates cherry-picked from @caseyprovost's PR

## Related work from the satellite gems

These PRs live in the now-absorbed repos and shaped what shipped here:
- graphiti-api/graphiti-rails#91 - rake task helpers out of the global namespace, landed as `Graphiti::Rails::RakeHelpers`
- graphiti-api/graphiti_spec_helpers#14 - RSpec matchers for attributes and relationships, reimplemented for 2.0 as `Graphiti::SpecHelpers::Matchers`
- graphiti-api/graphiti-rails#53 - a controller namespace option by @dineshpanda, superseded by `--controller`, which names the controller outright rather than deriving it from `.graphiticfg.yml`


sandstrom:
@jkeen Awesome!

I saw your comment in another issue but couldn't respond there because repo is archived.

Anyway, new docs looks great! 

A small piece of feedback:

Maybe just a bit more background in the first section?

> Graphiti is a serialization (and de-serialization) library for Ruby, with ready-made integrations for Rails.
> 
> It's tied to the [JSON:API](https://jsonapi.org/) spec, which means that your client layer (often a Javascript single page app) needs to speak the JSON API protocol.
> 
> The JSON API protocol itself is not complicated, so client-side logic can be hand rolled or you can use one of many libraries that are available.
> 
> This is an alternative to using for example JBuilder, which takes a more simple approach to JSON serialization.
> 
> Graphiti sits on top of your models and exposes them over a JSON:API-compliant interface. You define Resources instead of controllers and serializers, and get filtering, sorting, pagination, sparse fieldsets, statistics, and nested reads and writes across relationships, all over one endpoint.

This is just a quick stab as an example, but just trying to help people understand where this fits in.

jkeen:
@sandstrom Great feedback! I just updated them.

Let me know if you spot any other places that could be improved that would help de-mystify Graphiti. 

Also, fyi: the docs now live in this repo in plain markdown (in `/docs`, which docasaurus consumes from `/website`) and once this is merged to main those "Edit this page" links will work at the bottom of each page. I updated the readme on the 2.0/beta branch to point that out, and also put this intro suggestion on that readme as well.

sandstrom:
@jkeen Awesome!

Do you work at a company using Graphiti, or more of a personal involvement in the project?

In either case, this 2.0 update looks great and we've been looking for a replacement for fast-json-api (legacy gem) for a while. I'll take a look at this and try to make a switch for our rails app.

jkeen:
@sandstrom Good to hear! 

I've been using Graphiti professionally and personally since it was json-api-suite (2018) and when the original creators moved companies and dipped out of graphiti development I took over a couple of years ago. I was tip toeing around at first, but I made some big swings for this 2.0 release and I'm pretty dang thrilled with how it's shaping up. 

One thing that comes up a bunch from people who know and love Graphiti is "why isn't it more popular?" and I'm hoping some of these docs + ergonomic changes push it in the right direction. I feel like it solves a ton of problems, the tricky part I think has been explaining that clearly. 

Looking forward to hearing how your Graphiti transition goes, hit up the discord if you run into trouble.

factyy:
@jkeen , thanks for all the hard work towards this release! Awesome! :)

sandstrom:
Awesome! 💯

